PR #5, opened 2017-08-03
Validate Service name for Deploy functions endpoint

alexellis:
K8s won't allow a service name which doesn't conform to a DNS name. a-zA-Z and -

Changes:

* Introduce regex
* Add multi-stage build
* Add MIT license notice to Golang files


alexellis:
Appears to work when tested.

PR #9, opened 2017-08-06
Add a ServiceAccount to faas-netesd

luxas:
Make FaaS-netesd use the ServiceAccount the ClusterRoleBinding gives privileges to
Requires: #8

### DIFF
--- a/faas.yml
+++ b/faas.yml
@@ -14,6 +14,11 @@ spec:
   selector:
     app: faas-netesd
 ---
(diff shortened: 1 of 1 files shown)


alexellis:
Would the presence of these roles / values affect a cluster which doesn't have RBAC enabled?

luxas:
No. A ServiceAccount should be used regardless of authz method

PR #13, opened 2017-08-09
Use cluster-local DNS to reduce lookups

### LINKED ISSUE #12: Explicitly define the DNS service domain to reduce DNS lookups
Hi Alex,

Been working to run faas-netes as a demo ontop of [http://play-with-k8s.com](http://play-with-k8s.com)

While debugging some DNS failures, I noticed the faas gateway was swamping the DNS server with requests for possible permutations of a k8s or AWS service domain, ie;

<servicename>.default.svc.cluster.local
<servicename>.svc.cluster.local
<servicename>.cluster.local
<servicename>.ec2.internal

See [](https://gist.github.com/matjohn2/b182b80265bc581f93cd639def2c0792) for more tcpdump examples.

I thought setting `function_provider_url` in `faas.yml` to simply `faas-netesd` (the first name the lookups try) would solve this and not lookup any of the other domains, but it seems it's not just for finding faasd itself, the same hunting behaviour is also then seen with each configured function when hit via the gateway;

```
15:41:53.756886 IP (tos 0x0, ttl 64, id 34988, offset 0, flags [DF], proto UDP (17), length 181)
    10.32.0.3.domain > 10.44.0.0.46904: [bad udp cksum 0x1501 -> 0xb1b0!] 58910 NXDomain q: AAAA? urlping2.default.default.svc.cluster.local. 0/1/0 ns: cluster.local. SOA ns.dns.cluster.local. hostmaster.cluster.local. 1502290800 28800 7200 604800 60 (153)
```

Would it be possible to add an `env:` item into `faas.yml` such as:

`faas_service_lookup_domain` which in my case would be set to `default.svc.cluster.local, in order to prevent this try-em-all behaviour.

I ask specifically because PWK's has a default 150 concurrent DNS query limit, which this is dos-ing and i think PWK will create a nice testbed for teaching / using faas.

Thanks.

alexellis:

Fixes #12 

I'm not sure why Play-with-k8s has so many DNS search paths, but I've allowed an override for functions and updated the faas.yml file to use a more qualified cluster-local path.

### DIFF
--- a/faas.yml
+++ b/faas.yml
@@ -72,7 +72,7 @@ spec:
         imagePullPolicy: Always
         env:
         - name: functions_provider_url
-          value: "http://faas-netesd.default:8080/"
+          value: "http://faas-netesd.default.svc.cluster.local:8080/"
         ports:
         - containerPort: 8080
           protocol: TCP
--- a/handlers/proxy.go
+++ b/handlers/proxy.go
@@ -15,6 +15,8 @@ import (
 
 	"io/ioutil"
 
+	"os"
+
 	"github.com/gorilla/mux"
 )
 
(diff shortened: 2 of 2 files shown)


alexellis:
Putting this on hold temporarily.

alexellis:
Not sure this fixed the problem for @jockdarock and colleagues. Will mark revisit and close.

Issue #10, opened 2017-08-07
Create a Helm Chart for FaaS


errm:
A helm chart would make deploying FasS to Kubernetes much simpler, and it is quickly becoming the standard way to try out third party solutions on Kubernetes...

Once FaaS on kubernetes is somewhat stable we can then submit our chart to the official repo ... https://github.com/kubernetes/charts

I am happy to work on this unless someone else wants to give it a go...

alexellis:
Please go ahead @errm this sounds like a great idea. What repo would it live in?

errm:
I guess here for now ...

alexellis:
No problems. Does this also need #3 to be done?

alexellis:
Hey Ed have you had a go at this?

genexp:
We're actually doing this now.  We might be able to get a pull request in for it.  

cc: @cforbes

weikinhuang:
I opened a PR for this, can't use dynamic named services because a few of the containers have hard coded values.

alexellis:
@genexp @weikinhuang @errm what do we need to wrap this up? Does it test out OK?

@weikinhuang the hard-coded values are kept in the main FaaS repo here - https://github.com/alexellis/faas/tree/master/prometheus

^ that's where we need to update and then I have to re-push the builds from Docker Hub.

weikinhuang:
I'll leave it to everyone else on deciding if helm is going to be the preferred deployment method (which lets us get rid of having hard coded values in the main repo, and lets us use configmap mounted volumes). We can also keep the regular yaml files and update them with the configmaps that are found in the helm charts 
 (thereby removing them from the main repo as well and not having to build custom copies of prometheus and alertmanager).

alexellis:
I'd say we need to support both for now.. so let's go with the pain. It's easy to remove the hard-coded "vanilla" config in the future if helm is supported on all platforms we need.

weikinhuang:
I can create vanilla configmaps for this, would prevent you from having to build 2 images, and we can use stock docker hub images.

alexellis:
Derek add label: hacktoberfest

gardlt:
I would be very interested in a openfaas helm chart since i am currently using this for all my deployments. Let me know how I can help.

alexellis:
Thanks to @weikinhuang for leading this.

@genexp, @errm, @gardlt, @e-beach  if you can get involved with testing that would be great. I've compiled RBAC instructions here:

[...]

PR #22, opened 2017-09-12
Helm template

weikinhuang:
Signed-off-by: Wei Kin Huang <weikin.huang04@gmail.com>

Helm options:

```
functionNamespace=defaults to the deployed namespace, kube namespace to create function deployments in
async=true/false defaults to false, deploy nats if true
armhf=true/false, defaults to false, use arm images if true (missing images for async)
exposeServices=true/false, defaults to true, will always create `ClusterIP` services, and expose `NodePorts` if true
rbac=true/false defaults to true, if true create roles
```

### DIFF
--- a/helm/templates/rbac.yaml
+++ b/helm/templates/rbac.yaml
@@ -1,0 +1,80 @@
+{{- $functionNs := default .Release.Namespace .Values.functionNamespace }}
+{{- if .Values.rbac }}
+---
+apiVersion: rbac.authorization.k8s.io/v1beta1
+kind: ClusterRole
+metadata:
+  labels:
+    app: {{ template "faas-netesd.name" . }}
+    chart: {{ .Chart.Name }}-{{ .Chart.Version }}
+    component: faas-controller
+    heritage: {{ .Release.Service }}
+    release: {{ .Release.Name }}
+  name: faas-controller
+rules:
+  - apiGroups:
+      - ""
+    resources:
+      - services
+    verbs:
+      - get
+      - list
+      - watch
+      - create
+      - delete
+      - update
+  - apiGroups:
+    - extensions
+    resources:
+      - deployments
+    verbs:
+      - get
+      - list
+      - watch
+      - create
+      - delete
+      - update
+---
+apiVersion: rbac.authorization.k8s.io/v1beta1
+kind: ClusterRoleBinding
+metadata:
+  labels:
+    app: {{ template "faas-netesd.name" . }}
+    chart: {{ .Chart.Name }}-{{ .Chart.Version }}
+    component: faas-controller
+    heritage: {{ .Release.Service }}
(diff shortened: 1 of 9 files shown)


weikinhuang:
```bash
# install tiller
helm init --skip-refresh --upgrade

# for default
helm upgrade --install --debug --reset-values --set async=true openfaas helm/

# for non default ns/separate fn namespace
kubectl create ns openfaas
kubectl create ns openfaas-fn

helm upgrade --install --debug --namespace openfaas --reset-values --set async=true --set functionNamespace=openfaas-fn openfaas helm/
```

weikinhuang:
Refs #10

alexellis:
Very cool @weikinhuang !

weikinhuang:
@alexellis if the preferred method is to use helm, then I would suggest removing the vanilla yaml files and update the documentation. Otherwise you can keep both.

helm/values.yaml
```diff
@@ -0,0 +1,25 @@
...
+armhf: false
+exposeServices: true
+rbac: true
+
+# images
+images:
+  controller: weikinhuang/faas-netes:function-ns
+  controllerArmhf: functions/faas-netesd-armhf
+
+  gateway: functions/gateway:0.6.5
```
alexellis: We have 0.6.5 now too for AMD64

helm/templates/rbac.yaml
```diff
@@ -0,0 +1,80 @@
...
+    heritage: {{ .Release.Service }}
+    release: {{ .Release.Name }}
+  name: faas-controller
+rules:
+  - apiGroups:
+      - ""
+    resources:
+      - services
+    verbs:
+      - get
```
gardlt: we should try hosting this in the values file to keep the chart as configurable as possible.

weikinhuang: you can always bind more roles to the `faas-controller` user, this is the minimum permissions to get faas-netes working

alexellis:
I will merge the changes so we can start integrating them. We can update after some more testing / feedback.

@weikinhuang please add a section to the README.md with the notes in your PR here.

helm/templates/rbac.yaml
alexellis: @gardlt Can you help with testing?

alexellis:
@weikinhuang error during testing:

```
root@k8s-17-master:~/faas-netes# helm upgrade --install --debug --reset-values --set async=true openfaas helm/
[debug] Created tunnel using local port: '41016'

[debug] SERVER: "localhost:41016"

Error: UPGRADE FAILED: User "system:serviceaccount:kube-system:default" cannot list configmaps in the namespace "kube-system". (get configmaps)
```

weikinhuang:
@alexellis looks like your user does not have the right permissions

Issue #23, opened 2017-09-19
armhf (Raspberry Pi): AlertManager not ported yet


ecliptik:
## Expected Behaviour
Running

`kubectl apply -f ./faas.armhf.yml,monitoring.armhf.yml,rbac.yml` 

should bring up pods on Raspberry Pi k8s cluster.

## Current Behaviour
`alertmanager` pod godes into `CrashLoopBackOff`.

```
HypriotOS/armv7: pirate@navi in ~
$ kubectl get pods -o wide
NAME                            READY     STATUS             RESTARTS   AGE       IP           NODE
alertmanager-2609462557-bf54n   0/1       CrashLoopBackOff   142        11h       10.244.2.5   tatl
faas-netesd-1317931779-wk4gw    1/1       Running            0          11h       10.244.2.3   tatl
gateway-1085489343-cgs53        1/1       Running            0          11h       10.244.1.5   tael
prometheus-4259297277-qhj0b     1/1       Running            0          11h       10.244.2.4   tatl
```

Alertmanager pod log shows missing `/alertmanager.yml`

```
HypriotOS/armv7: pirate@navi in ~      
$ kubectl logs alertmanager-2609462557-bf54n
time="2017-09-19T18:50:52Z" level=info msg="Starting alertmanager (version=0.5.1, branch=master, revision=0ea1cac51e6a620ec09d053f0484b97932b5c902)" source="main.go:101" 
time="2017-09-19T18:50:52Z" level=info msg="Build context (go=go1.7.3, user=root@fb407787b8bf, date=20161125-08:19:00)" source="main.go:102" 
time="2017-09-19T18:50:52Z" level=info msg="Loading configuration file" file="/alertmanager.yml" source="main.go:195" 
time="2017-09-19T18:50:52Z" level=error msg="Loading configuration file failed: open /alertmanager.yml: no such file or directory" file="/alertmanager.yml" source="main.go:198"
```

## Possible Solution
Update https://github.com/alexellis/faas-netes/blob/master/monitoring.armhf.yml#L68 to 

`command: ["/bin/alertmanager","-config.file=/etc/alertmanager/config.yml", "-storage.path=/alertmanager"]`

## Steps to Reproduce (for bugs)
1. Setup Raspberry PI k8s cluster
2. `kubectl apply -f ./faas.armhf.yml,monitoring.armhf.yml,rbac.yml`

## Context
Following example doc: https://github.com/alexellis/faas/blob/master/guide/deployment_k8s.md

[...]

PR #24, opened 2017-09-19
Update armhf monitoring command to use proper path for config.yml - f…

ecliptik:
…ixes #23

### DIFF
--- a/monitoring.armhf.yml
+++ b/monitoring.armhf.yml
@@ -65,7 +65,7 @@ spec:
       - name: alertmanager
         image: alexellis2/alertmanager-armhf:0.5.1
         imagePullPolicy: Always
-        command: ["/bin/alertmanager","-config.file=/alertmanager.yml", "-storage.path=/alertmanager"]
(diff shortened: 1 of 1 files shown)


alexellis:
Thanks for looking at this. If you guys are wanting to use Kubernetes on ARM I'll prioritise sorting this out. The issue as I see it is that the image needs the config building or mounting in.

Example:

https://github.com/alexellis/faas/tree/master/prometheus

ecliptik:
@alexellis I wouldn't prioritize it, I just noticed it didn't work when following the how-to and figured out how to get it up at least to test things out on a Raspberry Pi cluster. Figured I should file an issue in case someone else came across it as well.

Looking at the Swarm config it does make sense to mount in a custom config to override a baked-in configuration. For now I can just re-build the image myself or use the default config.

alexellis:
I've fixed this by including something from @weikinhuang - https://github.com/openfaas/faas-netes/blob/master/monitoring-config.armhf.yml

Please apply this file. @ecliptik

PR #25, opened 2017-09-19
Added ACS yml with LB

tripdubroot:
Created and tested acs deployment located in:

/Cloud
  /Azure
    acs.yml

### DIFF
--- a/cloud/azure/acs.yml
+++ b/cloud/azure/acs.yml
@@ -1,0 +1,15 @@
+apiVersion: v1
+kind: Service
+metadata:
+  name: gateway
+  labels:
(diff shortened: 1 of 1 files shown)


alexellis:
Just curious - is there a way this could work where we maintain faas.yml and acs.yml is just the differences? I.e. people deploy with the node port then run this in after which switches to the LoadBalancer?

tripdubroot:
@alexellis I will test that out

alexellis:
Also as a note - Derek is picky so he has to get all commits signed off. Easiest way is to git reset and then commit again.

PR #31, opened 2017-09-29
we should explicitly choose a versioned one.

zouyee:
Signed-off-by: zou nengren <zouyee1989@gmail.com>
According to the comment for [Extensions](https://github.com/kubernetes/client-go/blob/master/kubernetes/clientset.go#L71-L73), it is deprecated, we should explicitly choose a versioned one.
@alexellis PTAL

### DIFF
--- a/handlers/delete.go
+++ b/handlers/delete.go
@@ -37,7 +37,7 @@ func MakeDeleteHandler(functionNamespace string, clientset *kubernetes.Clientset
 		getOpts := metav1.GetOptions{}
 
 		// This makes sure we don't delete non-labelled deployments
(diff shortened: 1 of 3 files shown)


alexellis:
Seems sensible @zouyee - can you sign-off and have you tested?

zouyee:
yes, this change is security.

alexellis:
How have you tested the changes? I can verify the PR early next week for merge

alexellis:
@zouyee please see above comment. cc @jockdarock (who may help with testing)

zouyee:
It has been tested!

alexellis:
Which version was it deprecated in? Is this safe for 1.6-1.8?

zouyee:
I do not the exact deadline,but picking a version is  recommended.

PR #37, opened 2017-10-06
feat(helm): adding-cpu-memory-resources

gardlt:
## Description

Adding Resource limitation on the deployments of the faas-netes

## Motivation and Context

## How Has This Been Tested?

I have deployed the helm chart and the kubernetes manifest in a minikube instance. 

## Types of changes

## Checklist:

### DIFF
--- a/openfaas/templates/faas.yaml
+++ b/openfaas/templates/faas.yaml
@@ -72,6 +72,10 @@ spec:
         image: {{ .Values.images.controller }}
         {{- end }}
         imagePullPolicy: Always
+        {{ if .Values.resources.enabled }}
+        resources:
+{{ toYaml .Values.resources.faasnetesd | indent 12 }}
+        {{- end }}
         env:
         - name: function_namespace
           value: {{ $functionNs | quote }}
@@ -124,6 +128,10 @@ spec:
         image: {{ .Values.images.gateway }}
         {{- end }}
         imagePullPolicy: Always
+        {{ if .Values.resources.enabled }}
+        resources:
+{{ toYaml .Values.resources.gateway | indent 12 }}
+        {{- end }}
         env:
         - name: functions_provider_url
           value: "http://faas-netesd.{{ .Release.Namespace }}:8080/"
--- a/openfaas/values.yaml
+++ b/openfaas/values.yaml
@@ -50,3 +50,47 @@ ingress:
     kubernetes.io/ingress.class: nginx
   tls:
     # Secrets must be manually created in the namespace.
+resources:
+  enabled: false
+  nats:
+    requests:
+      memory: "32MiB"
+      cpu: "50m"
+    limits:
+      memory: "32MiB"
+      cpu: "50m"
+  queueworker:
+    requests:
+      memory: "32MiB"
+      cpu: "50m"
+    limits:
+      memory: "32MiB"
+      cpu: "50m"
+  prometheus:
+    requests:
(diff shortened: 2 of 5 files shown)


alexellis:
Looks reasonable. I'd like to get a couple of people to look at the options.

rawkode:
Looks decent. Has there been any profiling of faas-netes to ensure the defaults we provide are mostly sufficient for common use-cases?

alexellis:
No, that is my concern by imposing a "finger in the air" limit.

weikinhuang:
This is definitely a good idea. It's usually best practice to have resource constraints so the scheduler doesn't create hotspots.

gardlt:
So the beauty of the helm chart is that you will be able to enable/disable the resources feature. Removing the "finger in the air" limit, I believe. For the hardcoded manifest, if we don't want it to impose we could remove...

weikinhuang:
If we can get average usage data for all the control plane components, we can easily add default resource requests and limits, and a user would be able to override it as they please.

weikinhuang:
Can we also instead change the format to something similar to the ones in the charts repository. Ex. https://github.com/kubernetes/charts/blob/master/stable/grafana/templates/deployment.yaml#L63

openfaas/templates/faas.yaml
```diff
@@ -72,6 +72,10 @@ spec:
         image: {{ .Values.images.controller }}
         {{- end }}
         imagePullPolicy: Always
+        {{ if .Values.resources.enabled }}
+        resources:
+{{ toYaml .Values.resources.faasnetesd | indent 12 }}
```
alexellis: @weikinhuang any thoughts for review?

alexellis:
The limits do seem high. The University of Calgary are running the swarm versions with 20mb of RAM as a limit for the gateway etc. @weikinhuang what is the best way to profile this? Would we get a good idea by running some kind of balanced load-test?

Can you guys work together on this via Slack? @weikinhuang @gardlt @rawkode we have the #faas-netes channel on Slack. :+1:

alexellis:
@gardlt

rawkode:
Based on my findings, I'd say 32MiB RAM and 50m CPU

Anyone going above that is probably going to be tweaking their values anyway.

alexellis:
I'm not sure if @gardlt is still working on this?

We now have PR https://github.com/openfaas/faas-netes/pull/73 from @resouer also

We need to add limits to force the correct QoS policy for the system-level pods.

alexellis:
@rawkode would you like to take over from @gardlt .. this PR seems to have been abandoned?

rawkode:
@alexellis Looks like you merged #37 instead?

alexellis:
This is #37 @rawkode

rawkode:
I meant #73 :joy:

openfaas/values.yaml
```diff
@@ -50,3 +50,47 @@ ingress:
...
+  queueworker:
+    requests:
+      memory: "32MiB"
+      cpu: "50m"
+    limits:
+      memory: "32MiB"
+      cpu: "50m"
+  prometheus:
+    requests:
+      memory: "32MiB"
```
alexellis: Needs 500MB RAM according to @stefanprodan.

stefanprodan: This is the RAM usage after 12h on GKE:
openfaas

PR #49, opened 2017-10-11
[Helm Chart] Create a Role instead of ClusterRole for functions in the same ns

weikinhuang:


alexellis:
Hi @weikinhuang can you provide a few details about the change?

ClusterRole vs Role etc.

weikinhuang:
You don't need to be a cluster admin to create a role I believe. Also there's no reason to create a cluster role if you're only accessing resources in a single namespace

PR #64, opened 2017-10-26
Helm: Fix ClusterRoleBinding for fns in different namespace

weikinhuang:
`ClusterRoleBinding` names are cluster wide unique, without this change, it would conflict with the control plane binding.


alexellis:
Thank you

Issue #72, opened 2017-11-01
OpenFaaS components should not be "BestEffort" pod


resouer:
The BestEffort pod (no resource definition) is the candidate to be evicted when node is under resource pressure. It is never recommended for system pods.
## Expected Behaviour

Faas components all have resource limit at least.
## Current Behaviour

Faas components all have no resource limit.
## Possible Solution

Add resource limit to Deployment.
## Steps to Reproduce (for bugs)

1. Create a kubernetes cluster on small size node
2. Enable pod eviction
3. faas-netes will be evicted to make room for other high priority pods

## Context

I have to patch faas-netes manually.

https://kubernetes.io/docs/tasks/configure-pod-container/quality-service-pod/#qos-classes

## Your Environment

* Docker version `docker version` (e.g. Docker 17.0.05 ):
Docker 1.12.6
* Are you using Docker Swarm or Kubernetes (FaaS-netes)?
Kubernetes 1.8.2 
* Operating System and version (e.g. Linux, Windows, MacOS):
ARM64 Linux
* Link to your project or a code example to reproduce issue:

alexellis:
Interesting.. we've not seen this yet. How small was the host you were using?

What memory usage did you observe for the faas-netesd controller?

resouer:
The host is 8Gi size. The memory usage is small, while t is evicted due to node lack of memory and faas-netes is a lowest QoS class pod.

alexellis:
What are you running on there to fill up 8GB? I run on a host with 1GB and I've not seen evictions with just very light use.

resouer:
The node is filled a lot of workloads, I should mention that  earlier :)

PR #73, opened 2017-11-01
Proposal: add memory limits to system-level pods

resouer:
Add memory limit to faas components.
## Motivation and Context

Do not use BestEffort pod to run faas components

Fixes: #72
Do not use `BestEffort` pod to run faas components. `BestEffort` pod will tend to be evicted.

See: https://kubernetes.io/docs/tasks/configure-pod-container/quality-service-pod/#qos-classes

## How Has This Been Tested?

Tested on ARM64 server on packet.net
Confirmed the available memory resource is 512Mi.

## Types of changes

## Checklist:

### DIFF
--- a/faas.arm64.yml
+++ b/faas.arm64.yml
@@ -38,6 +38,9 @@ spec:
         ports:
         - containerPort: 8080
           protocol: TCP
+        resources:
+          limits:
+            memory: 512Mi  
 ---
 apiVersion: v1
 kind: Service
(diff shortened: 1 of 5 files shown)


alexellis:
Seems valid. We've had a similar issue raised, but we're not sure what are good defaults. For the arm images 512mb is a lot of memory, maybe it should have a smaller limit for those.

Any changes to memory also need to be reflected in the helm charts.

resouer:
We can set it smaller, but one thing is: `limits` means it can use at most 512Mi, not claim to use 512Mi, the scheduler ignores `limits`.

alexellis:
This seems to be separate to the helm work and since this is a limit not a request I think we can merge, even if the number is very high.

PR #97, opened 2017-11-30
Set replica count on create/update from com.openfaas.scale.min label

### LINKED ISSUE #96: Feature: create function replica (Pod) count from metadata
To follow the style of Docker Swarm https://github.com/openfaas/faas/issues/414 we should update the create and update handlers so that if a label is specified for the minimum amount of replicas that is used instead of "1".

```
  labels:
    "com.openfaas.scale.min": "5"
    "com.openfaas.scale.max": "15"
```

In this instance we'd pick 5.

stefanprodan:
Fix #96 

Tested on Kubernetes 1.8 GKE

### DIFF
--- a/handlers/deploy.go
+++ b/handlers/deploy.go
@@ -147,13 +147,10 @@ func makeDeploymentSpec(request requests.CreateFunctionRequest, config *DeployHa
 	}
 
 	if request.Labels != nil {
+		if min := getMinReplicaCount(*request.Labels); min != nil {
+			initialReplicas = min
+		}
 		for k, v := range *request.Labels {
-			if k == "com.openfaas.scale.min" {
-				minReplicas, err := strconv.Atoi(v)
-				if err != nil && minReplicas > 0 {
-					initialReplicas = int32p(int32(minReplicas))
-				}
-			}
 			labels[k] = v
 		}
 	}
@@ -311,4 +308,17 @@ func createResources(request requests.CreateFunctionRequest) (*apiv1.ResourceReq
 	}
 
 	return resources, nil
+}
+
+func getMinReplicaCount(labels map[string]string) *int32 {
+	if value, exists := labels["com.openfaas.scale.min"]; exists {
+		minReplicas, err := strconv.Atoi(value)
+		if err != nil && minReplicas > 0 {
+			return int32p(int32(minReplicas))
+		} else {
+			log.Println(err)
+		}
+	}
+
+	return nil
 }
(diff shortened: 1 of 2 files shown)


handlers/deploy.go
```diff
@@ -139,19 +140,26 @@ func makeDeploymentSpec(request requests.CreateFunctionRequest, config *DeployHa
...
 	}
 
+	initialReplicas := int32p(initialReplicasCount)
+
 	labels := map[string]string{
 		"faas_function": request.Service,
 	}
+
 	if request.Labels != nil {
 		for k, v := range *request.Labels {
```
alexellis: There's a more idiomatic Golang pattern we can use here:
```
if value, exists := someMap[key]; exists {
 // do something with value
}
```

handlers/update.go
```diff
@@ -52,6 +53,12 @@ func MakeUpdateHandler(functionNamespace string, clientset *kubernetes.Clientset
 
 			if request.Labels != nil {
 				for k, v := range *request.Labels {
+					if k == "com.openfaas.scale.min" {
```
alexellis: This code can be extracted to a single common method.

handlers/deploy.go
```diff
@@ -139,19 +140,26 @@ func makeDeploymentSpec(request requests.CreateFunctionRequest, config *DeployHa
...
+			if k == "com.openfaas.scale.min" {
+				minReplicas, err := strconv.Atoi(v)
+				if err != nil && minReplicas > 0 {
```
alexellis: Let's at least log the error?

stefanprodan:
@alexellis the TravisCI build doesn't publish PRs to Docker Hub, I see errors in the build log:

```
Error parsing reference: "/faas-netesd:latest" is not a valid repository/tag: invalid reference format
Error response from daemon: Get https://registry-1.docker.io/v2/: unauthorized: incorrect username or password
invalid reference format
```

alexellis:
It shouldn't even be trying to push them to the hub.. can you checkout the .travis file and see what's different compared to the faas repo?

stefanprodan:
If it doesn't push PRs to Docker Hub how we suppose to test it? Each contributor would have to make repos in Docker hub for all openfaas images?

stefanprodan:
A smooth experience would be to push PRs to Docker Hub with a special tag like `pr97`, so if anyone wants to help with testing, he/she can easily pull the PR image.

stefanprodan:
@alexellis I've tested the PR with `stefanprodan/faas-netes:minpods2` image, after deployment the pod will scale to 5 is the label is present.

alexellis:
Thank you Stefan. This really helps!

Issue #136, opened 2018-02-12
Deployment with "rbac=false" is currently broken


hypnoglow:
Commit 44f7d42 broke "rbac=false" deployment mode, because now the ServiceAccount is only created when RBAC is enabled. 

## Expected Behaviour
openfaas helm chart successfully deployed.

## Current Behaviour
replicaset fails to create pods:
```
Events:
  Type     Reason        Age               From                   Message
  ----     ------        ----              ----                   -------
  Warning  FailedCreate  4m (x17 over 9m)  replicaset-controller  Error creating: pods "faas-netesd-655559f4fc-" is forbidden: error looking up service account openfaas/faas-controller: serviceaccount "faas-controller" not found
```

## Possible Solution
Solution 1: Bind pods to ServiceAccount only if RBAC is enabled.
Solution 2: Always create a ServiceAccount even if RBAC is disabled.

## Steps to Reproduce (for bugs)
1. `helm upgrade --install --debug --reset-values --set async=false --set rbac=false openfaas openfaas/`
2. `kubectl describe replicaset -l app=faas-netesd`

## Context
I've tried to deploy openfaas to my minikube cluster. But replicaset was stuck on creating pods.

## Your Environment

* Docker version `docker version` (e.g. Docker 17.0.05 ): 
```
ή docker version
Client:
 Version:	17.12.0-ce
 API version:	1.35
 Go version:	go1.9.2
 Git commit:	c97c6d6
 Built:	Wed Dec 27 20:03:51 2017
 OS/Arch:	darwin/amd64

Server:
 Engine:
  Version:	17.12.0-ce
  API version:	1.35 (minimum version 1.12)
  Go version:	go1.9.2
  Git commit:	c97c6d6
  Built:	Wed Dec 27 20:12:29 2017
  OS/Arch:	linux/amd64
  Experimental:	true
```

* Are you using Docker Swarm or Kubernetes (FaaS-netes)? 
Kubernetes (minikube)

* Operating System and version (e.g. Linux, Windows, MacOS):
macOS

* Link to your project or a code example to reproduce issue: 
\-

alexellis:
@rimusz

alexellis:
Hi @hypnoglow I don't think you've filled out the whole issue template. Please could you do that including which specific version of Kubernetes you are using and whether that is self-hosted or hosted on GKE, AKS etc etc?

hypnoglow:
@alexellis My bad, I thought it's unrelated. I've filled it.

[...]

PR #138, opened 2018-02-13
Ensure that the ServiceAccout is always created

LucasRoesler:
## Description
- Always create the ServiceAccount, regardless of the rbac setting
- Ensure that the ServiceAccount for the faasnetesd is always defined for the release namespace
- Ensure that the ServiceAccount is created for the `functionNamespace`, if it is different from the release namespace

## Motivation and Context

The is needed because some components assume the ServiceAccount exists, so it is easiest
to always create it.

Address #136 

## How Has This Been Tested?

**The default namespace**
Manually testing in minikube using 

```
helm install --values openfaas/values.yaml --set "rbac=false" ./openfaas
```

**A custom namespace**
```
kubectl create namespace openfaas
helm install --namespace openfaas --values openfaas/values.yaml --set "rbac=false" --set  ./openfaas
```

**Using two namespaces**
```
kubectl create namespace openfaas
kubectl create namespace minitest
helm install --namespace openfaas --values openfaas/values.yaml --set "rbac=false" --set "functionNamespace=minitest" ./openfaas
```

I then opened the web UI and deployed a function (certinfo) from the function store. 

Once the deploy finished, I was able to test the function and have it return the expected results.

## Types of changes

## Checklist:

### DIFF
--- a/chart/openfaas/templates/gateway.yaml
+++ b/chart/openfaas/templates/gateway.yaml
@@ -1,6 +1,18 @@
 {{- $functionNs := default .Release.Namespace .Values.functionNamespace }}
 ---
 apiVersion: v1
+kind: ServiceAccount
+metadata:
+  labels:
+    app: {{ template "openfaas.name" . }}
+    chart: {{ .Chart.Name }}-{{ .Chart.Version }}
+    component: faas-controller
+    heritage: {{ .Release.Service }}
+    release: {{ .Release.Name }}
+  name: faas-controller
+  namespace: {{ .Release.Namespace | quote }}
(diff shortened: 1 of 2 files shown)


alexellis:
Thank you for doing the more extensive testing on this change.

LucasRoesler:
I put the SA in the gateway file because it was previously in that file, I am not sure what the best location for it is.  I think either `rbac` file or `faasnetesd` would make sense.  I often think of SA as part of RBAC, so that could be good.  

It is easy to move around, we just need to pick where it goes.

PR #185, opened 2018-04-06
Add image_pull_policy configuration

### LINKED ISSUE #181: Proposal: allow imagePullPolicy to be set via configuration
## Description

This is a proposal from the community to allow the imagePullPolicy for functions to be set via configuration. The reason for this option is to support a developer workflow where images can be run as functions without pushing them to a repository. 

The current imagePullPolicy of "always" requires all images to be pushed to a registry - either a local or remote one. At least two people have asked for this change and as this supports a developer workflow and is a small change I think we should try to implement it.

I would expect us to introduce a new environmental variable - "image_pull_policy: always" for instance.

## Areas of change

* Code

https://github.com/openfaas/faas-netes/blob/ae7940a18548e31fe076e9ec2bb69721aa538b02/handlers/update.go

https://github.com/openfaas/faas-netes/blob/4723abe7e4e8cc8276e41042f7691e61ef0f7ff6/handlers/deploy.go

* Configuration changes

YAML and helm should be updated with an optional flag for helm for overriding the value

cc @stefanprodan for input and for a potential change in the faas-o6s CRD code too.

## Testing

This change can be tested by trying to deploy a function which has not been pushed remotely but exists within your Docker image library.

milsonian:
## Description
This adds an optional configuration via environment variable image_pull_policy which controls the Kubernetes ImagePullPolicy set on function deployments.
* Default remains 'Always'
* Can be overridden to 'IfNotPresent' or 'Never'
* Removed update support

Misc. notes:
* I took a guess at versioning in the chart, it will probably need an update.
* The (existing and new) tests could (should?) be moved near the packages they are testing, but have not addressed that here.
* There is a redundant import alias that could be cleaned up (`v1` and `apiv1`are the same), have not addressed but could add it here if desired or open a separate issue.

## Motivation and Context
 This change allows for simplified local development workflows when using minikube, and/or improved scalability with dynamic/CI-driven deployments and image tags.
Fixes #181

## How Has This Been Tested?
Tested in minikube locally with supported and unexpected values, deploying test/sample functions.  Verified `IfNotPresent` works with local images and remote on the same instance.

Added config handling to existing (basic) tests and ran those. Wired for string validation during config setup but kept it basic for now (it is handled only for explicitly known values downstream).

Add'l testing/validation would be appreciated!

## Types of changes

## Checklist:


chart/openfaas/values.yaml
```diff
@@ -19,7 +19,7 @@ queueWorker:
 
 # images
 images:
-  controller: functions/faas-netesd:0.4.6
+  controller: functions/faas-netesd:0.4.7
```
ericstoekl: 0.4.7 doesn't exist yet: https://hub.docker.com/r/functions/faas-netesd/tags/

types/read_config.go
```diff
@@ -71,4 +82,5 @@ type BootstrapConfig struct {
 	EnableFunctionReadinessProbe bool
 	ReadTimeout                  time.Duration
 	WriteTimeout                 time.Duration
+	ImagePullPolicy              string
```
ericstoekl: Why not make the `ImagePullPolicy` of type `apiv1.PullPolicy`? https://github.com/golang/build/blob/master/kubernetes/api/types.go#L863

ericstoekl:
Manually tested this and ran the included unit test -- it all looks good. Tested with an Ubuntu VM on AWS running kubectl directly (no minikube).

I made a few comments but overall it works well and does the minimum needed to implement this feature (which is very useful btw).

chart/openfaas/values.yaml
alexellis: Yes this is problematic - I can't merge the change due to this. Could you revert this part and then after I release do a new PR to helm / YAML?

handlers/deploy.go
```diff
@@ -156,6 +157,16 @@ func makeDeploymentSpec(request requests.CreateFunctionRequest, existingSecrets
 		return nil, resourceErr
 	}
 
+	var imagePullPolicy apiv1.PullPolicy
+	switch config.ImagePullPolicy {
+	case "Never":
```
alexellis: Just curious if if the value of `apiv1.PullNever` is the same as the string "Never" or whether it is an int?

handlers/update.go
```diff
@@ -43,7 +42,6 @@ func MakeUpdateHandler(functionNamespace string, clientset *kubernetes.Clientset
 
 		if len(deployment.Spec.Template.Spec.Containers) > 0 {
 			deployment.Spec.Template.Spec.Containers[0].Image = request.Image
-			deployment.Spec.Template.Spec.Containers[0].ImagePullPolicy = v1.PullAlways
```
alexellis: I'd rather comment this line out and state why it is left that way than remove it without the context.

server.go
```diff
@@ -41,9 +41,11 @@ func main() {
 	log.Printf("HTTP Read Timeout: %s\n", cfg.ReadTimeout)
 	log.Printf("HTTP Write Timeout: %s\n", cfg.WriteTimeout)
 	log.Printf("Function Readiness Probe Enabled: %v\n", cfg.EnableFunctionReadinessProbe)
+	log.Printf("Image Pull Policy: %s\n", cfg.ImagePullPolicy)
```
alexellis: Let's commit this for brevity.

alexellis:
Ping to @milsonian

milsonian:
Will take a look next week and patch it up, thanks.

chart/openfaas/values.yaml
milsonian: Yep I'll submit this separately.

types/read_config.go
milsonian: I left it generic here since `deploy.go` is where the deployment is being constructed (and the k8s api is already imported there); and because read_config (so far) is only using native go packages.  I could move it here but that is probably not necessary unless/until it's being reused by more than one handler.  If you guys feel strongly the logic should move here let me know.

handlers/deploy.go
milsonian: It ends up aligning with string constants in the k8s api, but I preferred to stick to the native types being used to construct the deploy object.

handlers/update.go
milsonian: I'm not a big fan of comments on deleted code - I could get more specific in the commit message.
Perhaps more importantly on this one I was thinking more - while the apispec claims it's not updatable, that's at the pod level; so in a deployment template, it probably is updatable since that creates a new replicaset.  It's worth deciding on-purpose however if we want it to be updatable in faasnetes, since it's effectively a global.  It could lead to unexpected behavior; consider e.g. in-place deployed/working functions that get mutated to `IfNotPresent` or `Never` (inadvertently or otherwise) after a faasnetes redeployment, which would break them.
I'm 50/50 on this one, thoughts?

server.go
milsonian: Not clear on what is desired here, can you be more specific?

milsonian:
Thanks for the feedback, replied to some comments, most importantly will need to know how we want to proceed on updates (see inline).

milsonian:
I squashed out the revert, and made the other requested changes.  Let me know if you'd like it all squashed.

milsonian:
Went ahead and squashed it all - @alexellis let me know if anything else is needed here, thanks

PR #192, opened 2018-04-25
Ensure that UpdateSecrets removes existing secrets before appending the new volume definitions

LucasRoesler:
## Description
Loop over the Volumes and VolumeMounts to remove the existing projected secrets volume definitions before appending the new volume and volume mount.

Without removing the existing volume information, you will have
duplicate volumes during a redeploy and this will cause an error in
kubernetes.

## Motivation and Context

Fixes openfaas/faas-cli#393

## How Has This Been Tested?
A new unit test has been added to verify that the volume definition is only added once.

## Types of changes

## Checklist:

- [X] I have added tests to cover my changes.

### DIFF
--- a/handlers/secrets.go
+++ b/handlers/secrets.go
@@ -83,13 +83,7 @@ func UpdateSecrets(request requests.CreateFunctionRequest, deployment *v1beta1.D
 
 	// remove the existing secrets volume, if we can find it. The update volume will be
 	// added below
-	existingVolumes := deployment.Spec.Template.Spec.Volumes
-	for i, v := range existingVolumes {
-		if v.Name == volumeName {
-			existingVolumes = append(existingVolumes[:i], existingVolumes[i+1:]...)
-		}
-	}
-
+	existingVolumes := removeVolume(volumeName, deployment.Spec.Template.Spec.Volumes)
 	deployment.Spec.Template.Spec.Volumes = existingVolumes
 	if len(secretVolumeProjections) > 0 {
 		deployment.Spec.Template.Spec.Volumes = append(existingVolumes, projectedSecrets)
@@ -105,16 +99,9 @@ func UpdateSecrets(request requests.CreateFunctionRequest, deployment *v1beta1.D
 		}
 
 		// remove the existing secrets volume mount, if we can find it. We update it later.
-		existingVolumeMounts := container.VolumeMounts
-		for i, v := range existingVolumeMounts {
-			if v.Name == volumeName {
-				existingVolumeMounts = append(existingVolumeMounts[:i], existingVolumeMounts[i+1:]...)
-			}
-		}
-
-		container.VolumeMounts = existingVolumeMounts
+		container.VolumeMounts = removeVolumeMount(volumeName, container.VolumeMounts)
 		if len(secretVolumeProjections) > 0 {
-			container.VolumeMounts = append(existingVolumeMounts, mount)
+			container.VolumeMounts = append(container.VolumeMounts, mount)
 		}
 
 		updatedContainers = append(updatedContainers, container)
@@ -123,4 +110,26 @@ func UpdateSecrets(request requests.CreateFunctionRequest, deployment *v1beta1.D
 	deployment.Spec.Template.Spec.Containers = updatedContainers
 
 	return nil
+}
+
+// removeVolume returns a Volume slice with any volumes matching volumeName removed
+func removeVolume(volumeName string, volumes []apiv1.Volume) []apiv1.Volume {
+	for i, v := range volumes {
+		if v.Name == volumeName {
+			volumes = append(volumes[:i], volumes[i+1:]...)
+		}
+	}
+
+	return volumes
+}
+
+// removeVolumeMount returns a VolumeMount slice with any mounts matching volumeName removed
+func removeVolumeMount(volumeName string, mounts []apiv1.VolumeMount) []apiv1.VolumeMount {
+	for i, v := range mounts {
+		if v.Name == volumeName {
+			mounts = append(mounts[:i], mounts[i+1:]...)
+		}
+	}
+
+	return mounts
 }
--- a/handlers/secrets.go
+++ b/handlers/secrets.go
@@ -112,24 +112,30 @@ func UpdateSecrets(request requests.CreateFunctionRequest, deployment *v1beta1.D
 	return nil
 }
 
-// removeVolume returns a Volume slice with any volumes matching volumeName removed
+// removeVolume returns a Volume slice with any volumes matching volumeName removed.
+// Uses the filter without allocation technique
+// https://github.com/golang/go/wiki/SliceTricks#filtering-without-allocating
 func removeVolume(volumeName string, volumes []apiv1.Volume) []apiv1.Volume {
-	for i, v := range volumes {
-		if v.Name == volumeName {
-			volumes = append(volumes[:i], volumes[i+1:]...)
+	newVolumes := volumes[:0]
+	for _, v := range volumes {
+		if v.Name != volumeName {
+			newVolumes = append(newVolumes, v)
 		}
 	}
 
-	return volumes
+	return newVolumes
 }
 
 // removeVolumeMount returns a VolumeMount slice with any mounts matching volumeName removed
+// Uses the filter without allocation technique
+// https://github.com/golang/go/wiki/SliceTricks#filtering-without-allocating
 func removeVolumeMount(volumeName string, mounts []apiv1.VolumeMount) []apiv1.VolumeMount {
-	for i, v := range mounts {
-		if v.Name == volumeName {
-			mounts = append(mounts[:i], mounts[i+1:]...)
+	newMounts := mounts[:0]
+	for _, v := range mounts {
+		if v.Name != volumeName {
+			newMounts = append(newMounts, v)
 		}
 	}
 
-	return mounts
+	return newMounts
 }
--- a/handlers/secrets.go
+++ b/handlers/secrets.go
@@ -71,47 +71,56 @@ func UpdateSecrets(request requests.CreateFunctionRequest, deployment *v1beta1.D
 		}
 	}
 
-	if len(secretVolumeProjections) > 0 {
-		volumeName := fmt.Sprintf("%s-projected-secrets", request.Service)
-		projectedSecrets := apiv1.Volume{
-			Name: volumeName,
-			VolumeSource: apiv1.VolumeSource{
-				Projected: &apiv1.ProjectedVolumeSource{
-					Sources: secretVolumeProjections,
-				},
+	volumeName := fmt.Sprintf("%s-projected-secrets", request.Service)
+	projectedSecrets := apiv1.Volume{
+		Name: volumeName,
+		VolumeSource: apiv1.VolumeSource{
+			Projected: &apiv1.ProjectedVolumeSource{
+				Sources: secretVolumeProjections,
 			},
-		}
-
-		existingVolumes := deployment.Spec.Template.Spec.Volumes
-		for i, v := range existingVolumes {
-			if v.Name == volumeName {
-				existingVolumes = append(existingVolumes[:i], existingVolumes[i+1:]...)
-			}
-		}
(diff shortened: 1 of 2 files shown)


LucasRoesler:
I am sure that the test could be streamlined and would love feedback on that

alexellis:
Lucas thank you for a quick turn-around. This may apply to the CRD via @stefanprodan but possibly not.

Quick question:

> Ensure that UpdateSecrets removes existing secrets before appending the new volume definitions 

What made you want to delete/attach over checking what was already there and keeping it?

LucasRoesler:
I figured that the new configuration is the preferred and desired configuration, so that should be the one that is used.  Also, if we are doing an update in k8s, it should be possible to trigger a roleback to the previous state, if desired.

handlers/secrets.go
```diff
@@ -82,7 +82,17 @@ func UpdateSecrets(request requests.CreateFunctionRequest, deployment *v1beta1.D
...
-		deployment.Spec.Template.Spec.Volumes = append(deployment.Spec.Template.Spec.Volumes, projectedSecrets)
+		existingVolumes := deployment.Spec.Template.Spec.Volumes
+		existingVolumeIndex := -1
+		for i, v := range existingVolumes {
+			if v.Name == volumeName {
+				existingVolumeIndex = i
+			}
+		}
+		if existingVolumeIndex > -1 {
+			existingVolumes = append(existingVolumes[:existingVolumeIndex], existingVolumes[existingVolumeIndex+1:]...)
```
sunilkumarmohanty: I think this can be moved to line number 89.

handlers/secrets.go
```diff
@@ -92,7 +102,19 @@ func UpdateSecrets(request requests.CreateFunctionRequest, deployment *v1beta1.D
...
+
+			existingVolumeMounts := container.VolumeMounts
+			existingVolumeMountIndex := -1
+			for i, v := range existingVolumeMounts {
+				if v.Name == volumeName {
+					existingVolumeMountIndex = i
+				}
+			}
+			if existingVolumeMountIndex > -1 {
+				existingVolumeMounts = append(existingVolumeMounts[:existingVolumeMountIndex], existingVolumeMounts[existingVolumeMountIndex+1:]...)
```
sunilkumarmohanty: This can be moved to line number 110.

handlers/secrets.go
LucasRoesler: I thought is was a faux-pas to edit the slice as your are iterating on it, but the tests still pass after I adjust it, so it doesn't seem to break.

alexellis: I need to find 10 minutes to sit down and review this in detail.

handlers/secrets.go
```diff
@@ -82,7 +82,13 @@ func UpdateSecrets(request requests.CreateFunctionRequest, deployment *v1beta1.D
 			},
 		}
 
-		deployment.Spec.Template.Spec.Volumes = append(deployment.Spec.Template.Spec.Volumes, projectedSecrets)
+		existingVolumes := deployment.Spec.Template.Spec.Volumes
+		for i, v := range existingVolumes {
```
alexellis: Could you break out a method?

LucasRoesler: Sorry if I am being dense, but I am not sure I entirely follow.  Which part of the `UpdateSecrets` function should be broken out further? Is there a portion of the code that should be reusable somewhere else?

alexellis:
@ryanfields we're putting a fix together for https://github.com/openfaas/faas-cli/issues/393

Please could you give this a shot on your environment?

handlers/secrets.go
```diff
@@ -82,7 +82,13 @@ func UpdateSecrets(request requests.CreateFunctionRequest, deployment *v1beta1.D
...
+			if v.Name == volumeName {
+				existingVolumes = append(existingVolumes[:i], existingVolumes[i+1:]...)
```
alexellis: Maybe I'm missing something.. why can't we use the more readable version?

```
existingVolumes = append(existingVolumes, v)
```

handlers/secrets_test.go
```diff
@@ -0,0 +1,130 @@
...
+
+import (
+	"testing"
+
+	"github.com/openfaas/faas/gateway/requests"
+	apiv1 "k8s.io/api/core/v1"
+	v1beta1 "k8s.io/api/extensions/v1beta1"
+)
+
+func Test_UpdateSecrets(t *testing.T) {
```
alexellis: This test name needs to be more descriptive so 6 months later we know what "Test Update Secrets" was trying to do.

handlers/secrets_test.go
```diff
@@ -0,0 +1,130 @@
...
+	if mount.Name != "testfunc-projected-secrets" {
+		t.Errorf("Incorrect volume mounts: expected \"testfunc-projected-secrets\", got \"%s\"", mount.Name)
+	}
+
+}
+
+func Test_UpdateSecrets_ReplacesPreviousSecretMount(t *testing.T) {
+	request := requests.CreateFunctionRequest{
+		Service: "testfunc",
+		Secrets: []string{"pullsecret", "testsecret"},
```
alexellis: What happens if `Secrets` is `nil`? Do we handle that gracefully?

handlers/secrets_test.go
alexellis: Also - I commented out the code change and this test still seemed to pass. I wasn't expecting that.

handlers/secrets.go
alexellis: Please could you change the style so that it's not destructive? I think readability is a concern too.

ryanfields:
@alexellis I just tried this out and I'm still getting the same error.  I manage openfaas with Helm.  Are there build steps I have to perform before using the chart that's in the fix branch?

alexellis:
How did you try it? Did you build a docker image and edit the helm chart?

ryanfields:
@alexellis I finally got it working.  Didn't notice that the build script was tagging the image as `functions/faas-netesd:latest` instead of `functions/faas-netesd:0.5.0`, so the helm upgrade was just using the release image instead of the one I'd built.

As far as I can tell, this fixes the issue.  I am able to call faas-cli deploy multiple times on functions that rely on secrets and they deploy with a 200 result.

handlers/secrets.go
johnmccabe: @alexellis its removing the element not appending it at this point, for example:
```
package main

import (
	"fmt"
)

func main() {
	s := []string{"a", "b", "c", "d", "e", "f"}
	fmt.Printf("original slice      : %s\n", s)
	for i, e := range s {
		if e == "c" {
			s = append(s[:i], s[i+1:]...)
		}
	}
	fmt.Printf("remove c from slice : %s\n", s)
}
```
https://play.golang.org/p/Yg6xixVtX44

johnmccabe: @LucasRoesler can  you `break` when you match the volume?

johnmccabe: @alexellis non-destructive alternative? (`Volume` needs pointed at the correct struct below)
```
existingVolumes := []Volume{}
for _, v := range  deployment.Spec.Template.Spec.Volumes{
    if v.Name != volumeName {
        existingVolumes = append(existingVolumes, v)
    }
}
```

LucasRoesler: @johnmccabe break definitely makes sense, if the Deployment is for an existing function coming from k8s, then the volume names are already unique and if it exists, it should only exist once.  I guess the question is, could there be a situation in which it was already duplicated in the volumes list? It would obviously be a broken deployment, but this code would cover that edge case and ensure that the volume really is only listed once.

LucasRoesler:
@alexellis and @johnmccabe  I have refactored the unit tests to have more verbose names and to be grouped a little bit. I have also expanded the test cases to cover the situation when a deploy has secrets and then a subsequent update wants to remove them, for completeness.  This change required a bit of a change in how the code flows.

handlers/secrets_test.go
alexellis: If this is meant to be a (retro-fitted) regression test then it should also have a better name describing what is being tested - `Test_UpdateSecrets` doesn't give enough information to someone having to maintain this code in the future.

handlers/secrets_test.go
```diff
@@ -0,0 +1,224 @@
...
+		err := UpdateSecrets(request, deployment, existingSecrets)
+		if err != nil {
+			t.Errorf("unexpected error %s", err.Error())
+		}
+
+		validateEmptySecretVolumesAndMounts(t, deployment)
+
+	})
+
+	t.Run("No volume added if request secrets is the empty slice", func(t *testing.T) {
```
alexellis: Why are these all sub-tests?

handlers/secrets.go
```diff
@@ -71,33 +71,56 @@ func UpdateSecrets(request requests.CreateFunctionRequest, deployment *v1beta1.D
...
 			},
+		},
+	}
+
+	// remove the existing secrets volume, if we can find it. The update volume will be
+	// added below
+	existingVolumes := deployment.Spec.Template.Spec.Volumes
+	for i, v := range existingVolumes {
+		if v.Name == volumeName {
+			existingVolumes = append(existingVolumes[:i], existingVolumes[i+1:]...)
```
alexellis: I think this code should be rewritten to express what it is doing. It's removing an element.

`removeVolume()` would do.

handlers/secrets_test.go
LucasRoesler: I was originally going to refactor out some of the common setup and possibly try to turn it into a proper table of tests

LucasRoesler: @alexellis  I just pushed an update that I think cleans up the test names

handlers/secrets.go
```diff
@@ -71,33 +71,65 @@ func UpdateSecrets(request requests.CreateFunctionRequest, deployment *v1beta1.D
...
+	deployment.Spec.Template.Spec.Containers = updatedContainers
+
 	return nil
 }
+
+// removeVolume returns a Volume slice with any volumes matching volumeName removed
+func removeVolume(volumeName string, volumes []apiv1.Volume) []apiv1.Volume {
+	for i, v := range volumes {
+		if v.Name == volumeName {
+			volumes = append(volumes[:i], volumes[i+1:]...)
```
alexellis: @johnmccabe left a comment about running a break at the point where you've matched?

handlers/secrets_test.go
```diff
@@ -0,0 +1,222 @@
...
+	err := UpdateSecrets(request, deployment, existingSecrets)
+	if err != nil {
+		t.Errorf("unexpected error %s", err.Error())
+	}
+
+	validateEmptySecretVolumesAndMounts(t, deployment)
+
+}
+
+func Test_UpdateSecrets_AddNewSecretVolume(t *testing.T) {
```
alexellis: Thanks for breaking out the tests.

handlers/secrets.go
LucasRoesler: Yes, I responded here https://github.com/openfaas/faas-netes/pull/192/files/b482effe30476916c17a20c17b436cab9e8600a3#r186285541

We can break the loop, but that will not ensure that there is exactly one instance of that volume in the list, it will only remove the first one found.  I think visiting the entire slice is the most accurate.

johnmccabe: Yeah this sounds ok.

handlers/secrets.go
LucasRoesler: Ok, easy enough, I will have to it tomorrow because I just left the house for the evening.

handlers/secrets.go
LucasRoesler: I added a new test to cover this edge case.  Adding the test revealed a bug in the filter implementation.  It will cause an index error because we are changing the length of the slice while iterating over it.  The new method is simpler and more standard filter method.  All tests are passing now.

alexellis:
Merged - I'll do a release.

stefanprodan:
@LucasRoesler can you please PR this change to https://github.com/openfaas-incubator/faas-o6s

LucasRoesler:
@stefanprodan the PR is open at openfaas-incubator/faas-o6s#16

PR #196, opened 2018-05-15
Get status of a single function directly

imikushin:
instead of listing ALL deployed functions and potentially iterating
over all of them.

Signed-off-by: Ivan Mikushin <imikushin@vmware.com>

### DIFF
--- a/handlers/reader.go
+++ b/handlers/reader.go
@@ -9,42 +9,59 @@ import (
 	"net/http"
 
 	"github.com/openfaas/faas/gateway/requests"
+	"k8s.io/api/apps/v1beta2"
+	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/kubernetes"
 )
 
-func getServiceList(functionNamespace string, clientset *kubernetes.Clientset) ([]requests.Function, error) {
-	functions := []requests.Function{}
+func getService(name, functionNamespace string, clientset *kubernetes.Clientset) (*requests.Function, error) {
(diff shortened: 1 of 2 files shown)


alexellis:
We're dual-maintaining two repos now.. could you also raise this for the operator if it applies? https://github.com/openfaas-incubator/faas-o6s

alexellis:
Ivan this seems like a useful contribution, however the contribution guide needs to be followed which means at the minimum filling out the full PR template. Please can you edit? https://raw.githubusercontent.com/openfaas/faas-netes/master/.github/PULL_REQUEST_TEMPLATE.md

Details of how you tested the change are also needed.

Thanks,

Alex

imikushin:
That's exactly the thing that got me stuck submitting a PR last time!

alexellis:
> Ivan this seems like a useful contribution, however the contribution guide needs to be followed which means at the minimum filling out the full PR template.

We have around 100 individual contributors now who have filled out the template and tested their changes.. 

> That's exactly the thing that got me stuck submitting a PR last time!

It'd be good to know what the sticking point is for you.

handlers/reader.go
```diff
@@ -9,42 +9,59 @@ import (
...
 	"k8s.io/client-go/kubernetes"
 )
 
-func getServiceList(functionNamespace string, clientset *kubernetes.Clientset) ([]requests.Function, error) {
-	functions := []requests.Function{}
+func getService(name, functionNamespace string, clientset *kubernetes.Clientset) (*requests.Function, error) {
+	deployment, err := clientset.AppsV1beta2().Deployments(functionNamespace).Get(name, metav1.GetOptions{})
+	if err != nil {
+		if errors.IsNotFound(err) {
+			return nil, nil
```
LucasRoesler: I am not sure we want to hide the not found error here, especially because the function does return an error, I would expect an error on the not found.

PR #212, opened 2018-06-04
Change ClusterRole/ClusterRoleBinding to Role/RoleBinding

### LINKED ISSUE #211: Changes needed to install in multiple k8s namespaces with Helm
I tried to install 2 openfaas envs in one k8s cluster's 2 different namespaces for different usage.
I'm using different helm release names, but the later installation will override cluster role and binding, hence the former env will not work.

## Expected Behaviour
It should be able to install openfaas seprately in multiple k8s namespaces

## Current Behaviour
The ClusterRole and ClusterRoleBinding is global across namespaces, and it's using a static name [faas-controller](https://github.com/openfaas/faas-netes/blob/master/chart/openfaas/templates/rbac.yaml#L27), this will cause overriding issue

## Possible Solution
Add helm release name as suffix to the name of ClusterRole/ClusterRoleBinding, so that different release will use different name, the permission will not be overrided.

1. install one release
```shell
helm upgrade --install openfaas openfaas/  \
  --namespace openfaas \
  --set functionNamespace=openfaas-fn
```
2. install another release
```shell
helm upgrade --install openfaas-dev openfaas/ \
   --namespace openfaas-dev \
   --set functionNamespace=openfaas-fn-dev
```

Then you'll find the first release will not work.

## Context
I'm setting up 2 envs for dev usage and prod usage

## Your Environment
k8s v1.9.2

martin-liu:
close #211 
Signed-off-by: Martin Liu <hflh1989@gmail.com>

## Description

~~Adding helm release name as suffix to names of ClusterRole and ClusterRoleBinding since they are global acrossing different namespaces, so that when install multiple openfaas in multiple namespaces, there'll not be overriding issue.~~

Change ClusterRole/ClusterRoleBinding to Role/RoleBinding, so that when install multiple openfaas in multiple namespaces, there'll not be overriding issue

Here's the detail:
1. Removed useless serviceaccount in fn namespace, since we only use the [serviceaccount in release namespace](https://github.com/openfaas/faas-netes/blob/master/chart/openfaas/templates/gateway-sa.yaml)
2. If `release namespace == fn namespace`, then we have role binding in the release namespace
3. if `release namespace != fn namespace`, then we have 2 role/rolebinding in both namespace and bind to the same `serviceaccount`

## Motivation and Context
fixes #211

## How Has This Been Tested?

I tested this in the following way:
  * Helm v2.8.2 + K8s cluster v1.9.2

## Types of changes

## Checklist:

### DIFF
--- a/chart/openfaas/Chart.yaml
+++ b/chart/openfaas/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v1
 description: Enable Kubernetes as a backend for OpenFaaS (Functions as a Service)
 name: openfaas
-version: 1.0.21
+version: 1.0.22
 sources:
 - https://github.com/openfaas/faas
 - https://github.com/openfaas/faas-netes
--- a/chart/openfaas/templates/rbac.yaml
+++ b/chart/openfaas/templates/rbac.yaml
@@ -1,30 +1,7 @@
 {{- $functionNs := default .Release.Namespace .Values.functionNamespace }}
-{{- if (ne .Release.Namespace $functionNs) }}
----
-apiVersion: v1
-kind: ServiceAccount
-metadata:
-  labels:
-    app: {{ template "openfaas.name" . }}
-    chart: {{ .Chart.Name }}-{{ .Chart.Version }}
-    component: faas-controller
-    heritage: {{ .Release.Service }}
-    release: {{ .Release.Name }}
-  name: faas-controller
-  namespace: {{ $functionNs | quote }}
-{{- end }}
-{{- if .Values.rbac }}
----
-apiVersion: rbac.authorization.k8s.io/v1beta1
-kind: ClusterRole
-metadata:
-  labels:
-    app: {{ template "openfaas.name" . }}
-    chart: {{ .Chart.Name }}-{{ .Chart.Version }}
-    component: faas-controller
-    heritage: {{ .Release.Service }}
-    release: {{ .Release.Name }}
-  name: faas-controller
+
+{{/* Generate role rules */}}
+{{ define "openfaas.role.rules" }}
 rules:
   - apiGroups:
       - ""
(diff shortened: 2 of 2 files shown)


alexellis:
@martin-liu what do you think about changing to Role/RoleBinding as per the definition in the ./yaml/ folder?

martin-liu:
@alexellis I think Role/RoleBinding is better, but I'm not familiar with the code, so I'm not sure if Role/RoleBinding can fit current requirement.

martin-liu:
@alexellis Updated code & description. Tested new code in my env(Helm v2.8.2 + K8s cluster v1.9.2)

alexellis:
I'll test this and then merge. Unless someone else from the community can help.

martin-liu:
@alexellis git rebased for Chart.yaml's version change

LucasRoesler:
I was able to do a quick test of deploying a function, worked as expected

alexellis:
I would still like to move this forward so in helm faas-netes only uses a Role/RoleBinding, but I think due to the latest helm chart changes we need @stefanprodan to give it a quick look over. The operator I believe will need a ClusterRole to be able to subscribe to events, but only read-only - the writing role for the operator is scoped to the openfaas-fn namespace.

https://blog.alexellis.io/introducing-the-openfaas-operator/

Alex

stefanprodan:
The operator runs under the `openfaas-operator` account that has it's own cluster role and permissions so this change will not impact the operator.

embano1:
Jumping on the discussion since I talked to @alexellis on Twitter about that.
+1 on removing `cluster` level bindings for the Helm chart (controller).

On the operator: since `events` are namespaced objects as well, the only reason (IMHO) for a cluster role would be to query x-namespace. Is that the case for the operator? Looks like it only works based on the CRD, which itself is a namespaced object, as per definition here: https://github.com/openfaas/faas-netes/blob/aa53feb8b16bc58c5945b03f3e653edf3a82ed2e/chart/openfaas/templates/crd.yaml#L13

alexellis:
Thanks for this PR, I believe we now have this for helm and YAML :+1:

PR #229, opened 2018-06-23
Refactor label parsing to set OF labels last

### LINKED ISSUE #209: Possible override of internal labels by the request labels
Currently we use the `faas_function` label as a selector in the Deployment of a function. But while debugging an issue with the selector, I noticed that we currently write all user supplied labels into the map that we supply to Kubernetes. 

Specifically, it looks like this

```
	labels := map[string]string{
		"faas_function": request.Service,
	}

	if request.Labels != nil {
		if min := getMinReplicaCount(*request.Labels); min != nil {
			initialReplicas = min
		}
		for k, v := range *request.Labels {
			labels[k] = v
		}
	}
```

With this [commit afdf2fc](https://github.com/openfaas/faas-netes/commit/afdf2fc7ea585c7b5a550e3c15c8fad7cfaa266d) we explicitly require that the `faas_function` label equal the service name.  If the user submits a different value for `faas_function`, it will break the deployment selector. 

## Expected Behaviour

The `faas_function` label should be strictly controlled by the `faas` system.

## Current Behaviour

The value can be overridden by the user.

## Possible Solution

We either need to skip select labels while processing the user input _or_ we need to ensure that we set the important internal labels after we have process the user labels. 

1. Create a stack.yaml with function `bar` and add the label `faas_function: foo`
2. Deploy the function
3. It should either generate an error about mismatching labels, for example
```
$ faas deploy
Deploying: testsecretapi.

Unexpected status: 500, message: Deployment.apps "testsecretapi" is invalid: spec.template.metadata.labels: Invalid value: map[string]string{"faas_function":"foo"}: `selector` does not match template `labels`
```

## Your Environment

* faas-netes version:  functions/faas-netesd:0.5.3
* Docker version `docker version` (e.g. Docker 17.0.05 ):  18.05.0-ce

* What version and distriubtion of Kubernetes are you using? `kubectl version`
```
Client Version: version.Info{Major:"1", Minor:"9", GitVersion:"v1.9.6", GitCommit:"9f8ebd171479bec0ada837d7ee641dec2f8c6dd1", GitTreeState:"clean", BuildDate:"2018-03-21T15:21:50Z", GoVersion:"go1.9.3", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"9", GitVersion:"v1.9.6", GitCommit:"9f8ebd171479bec0ada837d7ee641dec2f8c6dd1", GitTreeState:"clean", BuildDate:"2018-03-21T15:13:31Z", GoVersion:"go1.9.3", Compiler:"gc", Platform:"linux/amd64"}
```

LucasRoesler:
## Description
- Ensure that internal OF labels are set after user labels to ensure
that users do no override these internal values
- Refactor the getMinReplicaCount to work with the labels pointer, this
helps simplify the control flow in the handler methods
- Add constants for the label keys and update all references to those
keys throughout

## Motivation and Context

Closes #209

## How Has This Been Tested?

Unit tests have been added and I have tested in minikube

## Types of changes

## Checklist:


LucasRoesler:
Question for @alexellis  and @stefanprodan, while writing this I noticed that we are using pointers to maps in the request.  Concretely `request.Labels` is a `*map[string]string`, is there a specific reason for this?  Maps are already nil-able without the pointer. I did a couple experiments in the go playground 
 and the concrete map will still be nil if the field is omitted in the request json. So, unlike with strings, it doesn't seem that we need the pointer here to make the field optional.  It seems that we could just use a plain `map[string]string` instead of the pointer. I think this make some of the code a little more straightforward, since `*map[string]string` isn't indexable while `map[string]string` is.

alexellis:
There will be a Go-specific reason for using a reference here. Wouldn't changing it be a breaking change to the whole project?

Example of indexing: https://play.golang.org/p/cFdZ-Agh6ii

handlers/labels.go
```diff
@@ -0,0 +1,55 @@
...
+
+const (
+	// initialReplicasCount how many replicas to start of creating for a function, this is
+	// also used as the default return value for getMinReplicaCount
+	initialReplicasCount = 1
+
+	// OFFunctionNameLabel is the label key used by OpenFaaS to store the function name
+	// on the resources managed by OpenFaaS for that function.  This key is also used to
+	// denote that a resource is a "Function"
+	OFFunctionNameLabel = "faas_function"
```
alexellis: What does the `OF` prefix add?

handlers/labels.go
```diff
@@ -0,0 +1,55 @@
...
+	}
+
+	labels[OFFunctionNameLabel] = functionName
+
+	return labels
+}
+
+// getMinReplicaCount extracts the functions minimum replica count from the user's
+// request labels. If the value is not found, this will return the default value, 1.
+func getMinReplicaCount(labels *map[string]string) *int32 {
```
alexellis: It would be better to return an error than logging an error.

handlers/labels.go
```diff
@@ -0,0 +1,55 @@
...
+	// denote that a resource is a "Function"
+	OFFunctionNameLabel = "faas_function"
+	// OFFunctionMinReplicaCount is a label that user's can set and will be passed to Kubernetes
+	// as the Deployment replicas value.
+	OFFunctionMinReplicaCount = "com.openfaas.scale.min"
+)
+
+// parseLabels will copy the user request labels and ensure that any required internal labels
+// are set appropriately.
+func parseLabels(functionName string, requestLables *map[string]string) map[string]string {
```
alexellis: I'm not seeing any parsing in `parseLabels`. Is this the correct name for the function?

handlers/labels.go
LucasRoesler: In retrospect, those aren't needed.  I will remove them

handlers/labels.go
LucasRoesler: I will update it to return an error, I was simply moved the existing function.  The error printing was in the original function.

handlers/labels.go
LucasRoesler: Reasonable alternatives could be `buildLabels` or  `getLabels`

handlers/deploy.go
```diff
@@ -135,22 +131,9 @@ func makeDeploymentSpec(request requests.CreateFunctionRequest, existingSecrets
...
-		if min := getMinReplicaCount(*request.Labels); min != nil {
-			initialReplicas = min
-		}
-		for k, v := range *request.Labels {
-			labels[k] = v
-		}
-	}
-
+	initialReplicas := getMinReplicaCount(request.Labels)
+	labels := parseLabels(request.Service, request.Labels)
```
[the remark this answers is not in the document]
LucasRoesler: Sounds good, I originally intended to fold that into the function but wasn't very sure if since the logic only existed in the update

LucasRoesler:
I have pushed changes for each of your comments, they are individual commits to make it easier to review each change. I can squash the commits once it is approved.

Thinking more about how changing the `*map` would impact the project, you are right that it would breaking change, since each of the OF projects imports `requests` and references that struct. In theory it should just be a dependency update in each of the projects, but it is probably not worth the headache.

I was mostly curious from an academic perspective if there was a specific feature or capability that using a pointer there enabled?  It isn't obvious to me and I am just looking to learn more.

alexellis:
I need more time to understand this before merging.

PR #248, opened 2018-07-16
Set the ReadOnlyRootFilesystem option

LucasRoesler:
## Description
- Set the ReadOnlyRootFilesystem option in the SecurityContext of the
function
- Add a writable mount at `/tmp`
- Update the version of `faas` to use the new CreateFunctionRequest
struct

## Motivation and Context

This enforces that the function container is created with a read-only filesystem at `/`.

Please give all credit for the implementation to @Templum , I pulled in his diff but updated it with the new changes from `faas`.

Closes openfaas/faas#590 
Closes #228

## How Has This Been Tested?

I will be testing shortly once I have the PRs in place so that others can start taking a look.  The faas-cli companion PR is available in openfaas/faas-cli#439

## Types of changes

- [X] New feature (non-breaking change which adds functionality)

## Checklist:

- [X] My change requires a change to the documentation.

### DIFF
--- a/handlers/deploy.go
+++ b/handlers/deploy.go
@@ -248,24 +248,27 @@ func makeDeploymentSpec(request requests.CreateFunctionRequest, existingSecrets 
 							SecurityContext: &v1.SecurityContext{
 								ReadOnlyRootFilesystem: &request.ReadOnlyRootFilesystem,
 							},
-							VolumeMounts: []v1.VolumeMount{
-								{Name: "temp", MountPath: "/tmp", ReadOnly: false},
-							},
 						},
 					},
 					RestartPolicy: v1.RestartPolicyAlways,
 					DNSPolicy:     v1.DNSClusterFirst,
-					Volumes: []v1.Volume{
-						{
-							Name: "temp",
-							VolumeSource: v1.VolumeSource{
-								EmptyDir: &v1.EmptyDirVolumeSource{},
-							},
-						},
-					},
 				},
 			},
 		},
+	}
+
+	if request.ReadOnlyRootFilesystem {
+		deploymentSpec.Spec.Template.Spec.Containers[0].VolumeMounts = []v1.VolumeMount{
+			{Name: "temp", MountPath: "/tmp", ReadOnly: false},
+		}
+		deploymentSpec.Spec.Template.Spec.Volumes = []v1.Volume{
+			{
+				Name: "temp",
+				VolumeSource: v1.VolumeSource{
+					EmptyDir: &v1.EmptyDirVolumeSource{},
+				},
+			},
+		}
 	}
 
 	if err := UpdateSecrets(request, deploymentSpec, existingSecrets); err != nil {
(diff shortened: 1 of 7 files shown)


handlers/deploy.go
```diff
@@ -216,10 +216,24 @@ func makeDeploymentSpec(request requests.CreateFunctionRequest, existingSecrets
 							ImagePullPolicy: imagePullPolicy,
 							LivenessProbe:   probe,
 							ReadinessProbe:  probe,
+							SecurityContext: &v1.SecurityContext{
+								ReadOnlyRootFilesystem: &request.ReadOnlyRootFilesystem,
+							},
+							VolumeMounts: []v1.VolumeMount{
```
alexellis: Please create this if and only if the `request.ReadOnlyRootFilesystem` is set to true.

handlers/update.go
```diff
@@ -49,6 +49,7 @@ func MakeUpdateHandler(functionNamespace string, clientset *kubernetes.Clientset
 			//deployment.Spec.Template.Spec.Containers[0].ImagePullPolicy = v1.PullAlways
 
 			deployment.Spec.Template.Spec.Containers[0].Env = buildEnvVars(&request)
+			deployment.Spec.Template.Spec.Containers[0].SecurityContext.ReadOnlyRootFilesystem = &request.ReadOnlyRootFilesystem
```
alexellis: Suggest avoiding nil reference by checking if it exists first, creating if not `SecurityContext`

handlers/deploy.go
Templum: Sounds valid, I will adjust this

Templum: Well how do we achieve this, should I adjust it on my branch? And perform a rebase to have the matching History.

LucasRoesler:
I have tested this with my local cluster with a number of the store and sample functions.

 When `ReadOnlyRootFilesystem` is `true` the `/` directory and subfolders are read-only and the `tmp` folder is the only readable directory

When  `ReadOnlyRootFilesystem` is `false`, then the root is writable (if you are the root user). 

I have also tested switching the flag and updating a function both from `true` => `false` and back, `false` => `true`; each change updated the function as expected.

While testing I have definitely found that the CLI implementation does not work as expected.  I will have to dig into what is missing.  The CLI flag definitely works, but parsing from the stack file is not working. At this point, the gateway, swarm, and faasnetesd all seem to work as expected, the CLI is the only piece that I haven't verified as working.

alexellis:
@LucasRoesler this PR needs a rebase, I'm not sure what is different.

The code changes look relatively minor 👍 

Alex

alexellis:
> Please give all credit for the implementation to @Templum , I pulled in his diff but updated it with the new changes from faas.

Simon's PR showed how to pass the r/o flag, which was useful. Not to take away from that initial research, but you shouldn't do yourself a discredit here. You've finished the whole task including create/update/remove and unit testing.

Alex

alexellis:
Merged 👍

PR #249, opened 2018-07-16
Reduce the readiness checks for functions

### LINKED ISSUE #218: Use the /healthz endpoint to provide a more accurrate readiness check
The existing behavior of checking for the existence of /tmp/.lock does indeed verify that the container is up, but does not accurately validate the function is ready to receive requests.

The /healthz endpoint should provide a better mechanism for actual readiness

## Expected Behaviour

The kubernetes readiness and liveliness check should make an http request to the existing healtz endpoint of the function pod.

## Current Behaviour

The readiness and liveliness check simply check for the existence of a file /tmp/.lock

1.
2.
3.
4.

## Your Environment

* Docker version `docker version` (e.g. Docker 17.0.05 ):

* What version and distriubtion of Kubernetes are you using? `kubectl version`

* Operating System and version (e.g. Linux, Windows, MacOS):

* Link to your project or a code example to reproduce issue:

* What network driver are you using and what CIDR? i.e. Weave net / Flannel

berndtj:
* Significantly improves scale up time for functions (when going
  from 0 -> 1)
* Health check is hit more frequently, but should not noticibly
  impact performance
* Use the httpget probe type and leverage the watchdog /healthz
  endpoint

This could be optimized a little further if new image for doing
the http probes where created which would block on connection errors
and return immediately when the response comes back, but the best
case is < 1s improvement.

Some performance numbers.  Before (there was a timeout error):

cold start: 10.240251064300537
error calling function: Command 'echo -n "Test" | faas-cli -g http://192.168.64.78:31112 invoke hello-python' returned non-zero exit status 1.
cold start: 4.621361255645752
cold start: 5.6364970207214355
cold start: 11.648431777954102
cold start: 8.450724840164185
cold start: 9.854270935058594
cold start: 12.048357009887695
cold start: 12.24026870727539

After:

cold start: 1.8590199947357178
cold start: 1.8544681072235107
cold start: 2.065181016921997
cold start: 1.8414137363433838
cold start: 1.6598482131958008
cold start: 2.4577977657318115
cold start: 2.4510068893432617
cold start: 2.244048833847046
cold start: 2.6444039344787598

## Motivation and Context

Fix #218

## How Has This Been Tested?

```
import requests
import subprocess
import time

for i in range(10):
    now = time.time()
    try:
        subprocess.check_output('echo -n "Test" | faas-cli -g http://192.168.64.78:31112 invoke hello-python', shell=True)
        print("cold start: %s" % (time.time() - now))
    except Exception as e:
        print("error calling function: %s" % e)
    resp = requests.post("http://192.168.64.78:31113/system/scale-function/hello-python", json={"serviceName": "hello-python", "replicas": 0})
    time.sleep(10)
```

## Types of changes

This only breaks very old functions which use a version of the watchdog which does not have a /healtz endpoint

## Checklist:

### DIFF
--- a/chart/openfaas/templates/gateway-dep.yaml
+++ b/chart/openfaas/templates/gateway-dep.yaml
@@ -115,6 +115,20 @@ spec:
           value: "{{ .Values.faasnetesd.writeTimeout }}"
         - name: image_pull_policy
           value: {{ .Values.faasnetesd.imagePullPolicy | quote }}
+        - name: http_probe
+          value: "{{ .Values.faasnetesd.httpProbe }}"
+        - name: readiness_probe_initial_delay_seconds
+          value: "{{ .Values.faasnetesd.readinessProbe.initialDelaySeconds }}"
+        - name: readiness_probe_timeout_seconds
+          value: "{{ .Values.faasnetesd.readinessProbe.timeoutSeconds }}"
+        - name: readiness_probe_period_seconds
+          value: "{{ .Values.faasnetesd.readinessProbe.periodSeconds }}"
+        - name: liveness_probe_initial_delay_seconds
+          value: "{{ .Values.faasnetesd.livenessProbe.initialDelaySeconds }}"
+        - name: liveness_probe_timeout_seconds
+          value: "{{ .Values.faasnetesd.livenessProbe.timeoutSeconds }}"
+        - name: liveness_probe_period_seconds
+          value: "{{ .Values.faasnetesd.livenessProbe.periodSeconds }}"
         ports:
         - containerPort: 8081
           protocol: TCP
--- a/handlers/deploy.go
+++ b/handlers/deploy.go
@@ -44,9 +44,18 @@ func ValidateDeployRequest(request *requests.CreateFunctionRequest) error {
 	return fmt.Errorf("(%s) must be a valid DNS entry for service name", request.Service)
 }
 
+// FunctionProbeConfig specify options for Liveliness and Readiness checks
+type FunctionProbeConfig struct {
+	InitialDelaySeconds int32
+	TimeoutSeconds      int32
+	PeriodSeconds       int32
+}
+
 // DeployHandlerConfig specify options for Deployments
 type DeployHandlerConfig struct {
-	EnableFunctionReadinessProbe bool
+	HTTPProbe                    bool
+	FunctionReadinessProbeConfig *FunctionProbeConfig
+	FunctionLivenessProbeConfig  *FunctionProbeConfig
 	ImagePullPolicy              string
 }
 
@@ -118,21 +127,41 @@ func MakeDeployHandler(functionNamespace string, clientset *kubernetes.Clientset
 
 func makeDeploymentSpec(request requests.CreateFunctionRequest, existingSecrets map[string]*apiv1.Secret, config *DeployHandlerConfig) (*v1beta1.Deployment, error) {
 	envVars := buildEnvVars(&request)
-	path := filepath.Join(os.TempDir(), ".lock")
-	probe := &apiv1.Probe{
-		Handler: apiv1.Handler{
+	var handler apiv1.Handler
+
+	if config.HTTPProbe {
+		handler = apiv1.Handler{
+			HTTPGet: &apiv1.HTTPGetAction{
+				Path: "/_/health",
+				Port: intstr.IntOrString{
+					Type:   intstr.Int,
+					IntVal: int32(watchdogPort),
+				},
+			},
+		}
+	} else {
+		path := filepath.Join(os.TempDir(), ".lock")
+		handler = apiv1.Handler{
 			Exec: &apiv1.ExecAction{
 				Command: []string{"cat", path},
 			},
-		},
-		InitialDelaySeconds: 3,
-		TimeoutSeconds:      1,
-		PeriodSeconds:       10,
+		}
+	}
+	readinessProbe := &apiv1.Probe{
+		Handler:             handler,
+		InitialDelaySeconds: config.FunctionReadinessProbeConfig.InitialDelaySeconds,
+		TimeoutSeconds:      config.FunctionReadinessProbeConfig.TimeoutSeconds,
+		PeriodSeconds:       config.FunctionReadinessProbeConfig.PeriodSeconds,
 		SuccessThreshold:    1,
 		FailureThreshold:    3,
 	}
-	if !config.EnableFunctionReadinessProbe {
-		probe = nil
+	livenessProbe := &apiv1.Probe{
+		Handler:             handler,
+		InitialDelaySeconds: config.FunctionLivenessProbeConfig.InitialDelaySeconds,
+		TimeoutSeconds:      config.FunctionLivenessProbeConfig.TimeoutSeconds,
+		PeriodSeconds:       config.FunctionLivenessProbeConfig.PeriodSeconds,
+		SuccessThreshold:    1,
+		FailureThreshold:    3,
 	}
 
 	initialReplicas := int32p(initialReplicasCount)
@@ -214,8 +243,8 @@ func makeDeploymentSpec(request requests.CreateFunctionRequest, existingSecrets 
 							Env:             envVars,
 							Resources:       *resources,
 							ImagePullPolicy: imagePullPolicy,
-							LivenessProbe:   probe,
-							ReadinessProbe:  probe,
+							LivenessProbe:   livenessProbe,
+							ReadinessProbe:  readinessProbe,
 						},
 					},
 					RestartPolicy: v1.RestartPolicyAlways,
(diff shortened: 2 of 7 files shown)


handlers/deploy.go
```diff
@@ -118,16 +116,19 @@ func MakeDeployHandler(functionNamespace string, clientset *kubernetes.Clientset
 
 func makeDeploymentSpec(request requests.CreateFunctionRequest, existingSecrets map[string]*apiv1.Secret, config *DeployHandlerConfig) (*v1beta1.Deployment, error) {
 	envVars := buildEnvVars(&request)
-	path := filepath.Join(os.TempDir(), ".lock")
 	probe := &apiv1.Probe{
 		Handler: apiv1.Handler{
-			Exec: &apiv1.ExecAction{
-				Command: []string{"cat", path},
+			HTTPGet: &apiv1.HTTPGetAction{
```
alexellis: This can't be turned on by default, it needs to be optional.

handlers/deploy.go
```diff
@@ -118,16 +116,19 @@ func MakeDeployHandler(functionNamespace string, clientset *kubernetes.Clientset
...
+				Path: "/healthz",
+				Port: intstr.IntOrString{
+					Type:   intstr.Int,
+					IntVal: int32(watchdogPort),
+				},
 			},
 		},
-		InitialDelaySeconds: 3,
+		InitialDelaySeconds: 0,
```
alexellis: Please introduce a configuration item for this. You can largely copy and paste from the existing variables.

Should also be available via helm as an option.

handlers/deploy.go
```diff
@@ -118,16 +116,19 @@ func MakeDeployHandler(functionNamespace string, clientset *kubernetes.Clientset
...
 		TimeoutSeconds:      1,
-		PeriodSeconds:       10,
+		PeriodSeconds:       1,
```
alexellis: This should also be a configuration item with a default of the previous value for compatibility. When used in dispatch you'd just set your values via the helm chart

handlers/deploy.go
berndtj: Oh, I had thought based on our off-line conversation this wasn't the case.  Easy to make optional

handlers/deploy.go
berndtj: ok

handlers/deploy.go
berndtj: Are you sure you want the default to be the previous value(s)?  These values will have far more positive effect than negative, and shouldn't break anything existing

berndtj:
Yes @alexellis I assumed any change to the actual probe, would be a separate PR

berndtj:
Pretty much exposed everything.  Let me know if you think this is going a bit far.  Also, I left the default for the liveness probe the same as before, but the readiness probe has new values which make 0->1 scaling faster.

alexellis:
Hi @berndtj that is very thorough work, thanks for taking time to think through the configuration options and for signing-off the PR.

Here is what I was thinking:

Since we use the same-point for liveness and readiness, they should always be enabled, but the question is which mode. Compatibility mode or http-mode?

```
   probe_type: http 
```

```
   probe_type: lock 
```

Both are needed to keep compatibility with existing functions, that's why the option is needed.

Given a value of `http` then the `/_/health` endpoint should be queried (as defined in the watchdog). The OpenFaaS watchdog uses a prefix to avoid any clashing of function endpoints:

```
/_/health
```

https://github.com/openfaas/faas/blob/master/watchdog/main.go#L53

faas-netes and the gateway expose health via `/healthz` because they are not functions, but services.

If you are not using the watchdog and don't want to expose your health endpoint via `/_/health` then perhaps this should be configurable in the helm chart, for your use only?

Given a value of `lock` then the existing code should still run and the `http` probe should not be added.

I think we could de-duplicate the options in the PR and use the same values for timeout / period checking and initial check for both liveness and readiness. At this stage they point at the same endpoint and react in the same way.

What are your thoughts on above?

berndtj:
That's kind of embarrassing (/healthz vs /_/health).  I'm surprised it still passes readiness/liveness.  I'm not sure the value needs to be configurable necessarily.

Anyway on to your other points.  Yeah, I forgot about "compatiblity" mode, that's easy enough.

I explicitly did not dedupe the probes as I figured you actually want different values for liveness and readiness even if the endpoint is the same.  For instance, I can live with a much longer period with liveness, but I want as short as possible for readiness.

berndtj:
Yes, I only see it scaling from 0->1 (but to be honest I'm not testing subsequent requests).  We've also seen it with Dispatch and openfaas when we are waiting on the function to become ready.  It's likely the same issue.

I actually have a change to the actual readiness check that doesn't rely on the lock file at all.  I'll give that a test.  I think it actually does fix the issue.

types/read_config.go
```diff
@@ -92,9 +108,15 @@ func (ReadConfig) Read(hasEnv HasEnv) BootstrapConfig {
 
 // BootstrapConfig for the process.
 type BootstrapConfig struct {
-	EnableFunctionReadinessProbe bool
-	ReadTimeout                  time.Duration
-	WriteTimeout                 time.Duration
-	ImagePullPolicy              string
-	Port                         int
+	HTTPProbe                         bool
```
alexellis: Think this might be useful to comment on:

```
// HTTPProbe when set to true switches readiness and liveness probe to access /_/health over HTTP instead of accessing /tmp/.lock.
```

types/read_config.go
```diff
@@ -92,9 +108,15 @@ func (ReadConfig) Read(hasEnv HasEnv) BootstrapConfig {
...
+	ReadinessProbeInitialDelaySeconds int
+	ReadinessProbeTimeoutSeconds      int
+	ReadinessProbePeriodSeconds       int
```
alexellis: Curious if this is worth making a Golang duration in this PR or a follow-up?

The other configs support Golang durations now, could call `durationVal.Seconds()` in the code to convert if that makes sense.

chart/openfaas/templates/gateway-dep.yaml
```diff
@@ -113,6 +113,20 @@ spec:
           value: "{{ .Values.faasnetesd.writeTimeout }}"
         - name: image_pull_policy
           value: {{ .Values.faasnetesd.imagePullPolicy | quote }}
+        - name: http_probe
```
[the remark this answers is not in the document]
alexellis: (Just seen the defaults in the code, if the defaults work well then we could update the YAML later.) Best way to test is to `kubectl delete` the two OpenFaaS namespaces, then apply the YAML folder again.

types/read_config.go
alexellis: I don't consider this as compulsory - just want your take on it.

berndtj:
I hadn't even considered the yaml ;).  I'll make sure and test first

handlers/deploy.go
```diff
@@ -118,21 +127,41 @@ func MakeDeployHandler(functionNamespace string, clientset *kubernetes.Clientset
 
 func makeDeploymentSpec(request requests.CreateFunctionRequest, existingSecrets map[string]*apiv1.Secret, config *DeployHandlerConfig) (*v1beta1.Deployment, error) {
 	envVars := buildEnvVars(&request)
-	path := filepath.Join(os.TempDir(), ".lock")
-	probe := &apiv1.Probe{
-		Handler: apiv1.Handler{
+	var handler apiv1.Handler
+
+	if config.HTTPProbe {
```
stefanprodan: This looks good for now but in the future we should have a way to switch this flag from the function definition so that is backwards compatible with functions that are built with the old watchdog.

types/read_config.go
```diff
@@ -92,9 +108,15 @@ func (ReadConfig) Read(hasEnv HasEnv) BootstrapConfig {
...
+	LivenessProbeInitialDelaySeconds  int
+	LivenessProbeTimeoutSeconds       int
+	LivenessProbePeriodSeconds        int
```
stefanprodan: I would make these of type `time.Duration` but we can address this at a later time.

handlers/deploy.go
alexellis: For instance, we could use the new annotations field being worked on by @ewilde

alexellis:
Not going to be popular for saying this, but we've had some Chart changes merged since the PR.

This generally means resetting the commit, rebasing the chart then running make charts again before doing a commit with a force.

Other than that LGTM.

Alex

alexellis:
Hi Berndt I know you have time away coming up, all I could do at this point is to take your commit, reset it, fix it and add it back again but it would lose your authorship. I could perhaps set the "git author" but it won't look like it does now in the history.

Alex

berndtj:
I can fix it up right now.

alexellis:
We just discovered the tests were broken in this commit, fixed in #249.

dkozlov:
Hi @berndtj,  Do you have plans to add /_/health endpoint in https://github.com/openfaas-incubator/of-watchdog functions?

alexellis:
This is the wrong repo for the question. There's already a disk-based health check with s http one in progress - https://github.com/openfaas-incubator/of-watchdog/pull/13

PR #258, opened 2018-07-22
Add annotation support

ewilde:
## Description
Store stack.yml annotation meta data in k8s service and deployment
specifications

## Motivation and Context
It is more appropriate in kubernetes to store meta data in the annotations attribute and not labels. Labels have a more restricted value range and are not designed for general purpose metadata see: https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/

## How Has This Been Tested?

1. Deploy openfaas using minikube and latest versions of `gateway` (currently https://github.com/openfaas/faas/commit/4cbb7d968d21f4ae2daacfb27213c9371d4d4449)

**Using a Faas-cli supporting annotations**, a patched version of the [faas-cli that supports annotation](https://github.com/ewilde/faas-cli/tree/annotations):

Using the following `stack.yml`
```yaml
provider:
  name: faas
  gateway: http://192.168.99.100:31112

functions:
  nodejs-echo:
    lang: node
    handler: ./sample/nodejs-echo
    image: alexellis/faas-nodejs-echo:0.1
    labels:
      foo: "bar"
    annotations:
      current.time: "first-deploy"
```

1. Create a new deployment
`faas-cli deploy -f ./stack.yml`

image

Annotations were visible on the minikube dashboard for the following resources
* Deployment (yes)

```
 kubectl --namespace=openfaas-fn get deployment  -o json | jq '.items[].metadata.annotations'

{
  "current.time": "first-deploy",
  "deployment.kubernetes.io/revision": "1",
  "prometheus.io.scrape": "false"
}
```

* Pods (yes)

```
$ kubectl --namespace=openfaas-fn get pod  -o json | jq '.items[].metadata.annotations

{
  "current.time": "first-deploy",
  "prometheus.io.scrape": "false"
}
```

* Replica sets (yes)

```
kubectl --namespace=openfaas-fn get replicaset  -o json | jq '.items[].metadata.annotations'

{
  "current.time": "first-deploy",
  "deployment.kubernetes.io/desired-replicas": "1",
  "deployment.kubernetes.io/max-replicas": "2",
  "deployment.kubernetes.io/revision": "1",
  "prometheus.io.scrape": "false"
}
```

* Services (yes)

```
 kubectl --namespace=openfaas-fn get service  -o json | jq '.items[].metadata.annotations'

{
  "current.time": "first-deploy",
  "prometheus.io.scrape": "false"
}
```

2. Updating an existing function and changing the annotation

Example yaml used in test:
```
functions:
  nodejs-echo:
    lang: node
    handler: ./sample/nodejs-echo
    image: alexellis/faas-nodejs-echo:0.1
    labels:
      foo: "bar"
    annotations:
      current.time: "update-1"
```
```
$ faas-cli deploy -f ./stack.yml
Deploying: nodejs-echo.
Function nodejs-echo already exists, attempting rolling-update.

Deployed. 200 OK.
URL: http://192.168.99.100:31112/function/nodejs-echo

```

 * Deployment (yes)

```
kubectl --namespace=openfaas-fn get deployment  -o json | jq '.items[].metadata.annotations'

{
  "current.time": "update-1",
  "deployment.kubernetes.io/revision": "2",
  "prometheus.io.scrape": "false"
}
```

 * Pods (yes)

```
kubectl --namespace=openfaas-fn get pod  -o json | jq '.items[].metadata.annotations'

{
  "current.time": "update-1",
  "prometheus.io.scrape": "false"
}
```
 * Replica sets (yes)

```
kubectl --namespace=openfaas-fn get replicaset  -o json | jq '.items[1].metadata.annotations'

{
  "current.time": "update-1",
  "deployment.kubernetes.io/desired-replicas": "1",
  "deployment.kubernetes.io/max-replicas": "2",
  "deployment.kubernetes.io/revision": "2",
  "prometheus.io.scrape": "false"
}
```

 * Services (no)
```
kubectl --namespace=openfaas-fn get service  -o json | jq '.items[].metadata.annotations'

{
  "current.time": "first-deploy",
  "prometheus.io.scrape": "false"
}
```

3. Update function and add a new annotation
 * Deployment (yes)
 * Pods (yes)
 * Replica sets  (yes)
 * Services (no)

Just to contrast the above if we do this with a label we get:

4. Update with a new label
 * Deployment no
 * Pods (yes)
 * Replica sets (yes)
 * Services (n/a)

Note updating an existing label doesn't seem to work at the moment

5. Updating annotation using `kubectl`

```
$ kubectl  --namespace=openfaas-fn annotate  --overwrite deployment nodejs-echo current.time=kubectl%time
deployment.extensions/nodejs-echo annotated
```
* Deployment (yes)

```
 kubectl --namespace=openfaas-fn get deployment  -o json | jq '.items[].metadata.annotations'

{
  "current.time": "kubectl%time",
  "deployment.kubernetes.io/revision": "3",
  "next.time": "first-1",
  "prometheus.io.scrape": "false"
}
```

* Pods (No)

```
$ kubectl --namespace=openfaas-fn get pod  -o json | jq '.items[].metadata.annotations'

{
  "current.time": "update-1",
  "next.time": "first-1",
  "prometheus.io.scrape": "false"
}
```

* Replica sets (yes)

```
kubectl --namespace=openfaas-fn get replicaset  -o json | jq '.items[].metadata.annotations'

{
  "current.time": "kubectl%time",
  "deployment.kubernetes.io/desired-replicas": "1",
  "deployment.kubernetes.io/max-replicas": "2",
  "deployment.kubernetes.io/revision": "3",
  "deployment.kubernetes.io/revision-history": "3,3",
  "next.time": "first-1",
  "prometheus.io.scrape": "false"
}
```

* Services (No)

```
 kubectl --namespace=openfaas-fn get service  -o json | jq '.items[].metadata.annotations'

{
  "current.time": "first-deploy",
  "prometheus.io.scrape": "false"
}
```

**Using a version of the faas-cli that doesn't support annotations, but faas-netes does**

Works as before:

image

**Using a version of the faas-cli and gateway that doesn't support annotations, but faas-netes does**

Works as before:

image

## Types of changes

## Checklist:

- [] My change requires a change to the documentation.

### DIFF
--- a/Gopkg.toml
+++ b/Gopkg.toml
@@ -5,7 +5,7 @@
 
 [[constraint]]
   name = "github.com/openfaas/faas"
-  version = "0.8.6"
+  version = "0.8.8"
 
 [[constraint]]
   name = "github.com/openfaas/faas-provider"
--- a/handlers/deploy.go
+++ b/handlers/deploy.go
@@ -196,6 +196,7 @@ func makeDeploymentSpec(request requests.CreateFunctionRequest, existingSecrets 
 		imagePullPolicy = apiv1.PullAlways
 	}
 
+	annotations := buildAnnotations(request)
 	deploymentSpec := &v1beta1.Deployment{
 		TypeMeta: metav1.TypeMeta{
 			Kind:       "Deployment",
@@ -203,6 +204,7 @@ func makeDeploymentSpec(request requests.CreateFunctionRequest, existingSecrets 
 		},
 		ObjectMeta: metav1.ObjectMeta{
 			Name: request.Service,
+			Annotations: annotations,
 		},
 		Spec: v1beta1.DeploymentSpec{
 			Selector: &metav1.LabelSelector{
@@ -229,7 +231,7 @@ func makeDeploymentSpec(request requests.CreateFunctionRequest, existingSecrets 
 				ObjectMeta: metav1.ObjectMeta{
 					Name:        request.Service,
 					Labels:      labels,
-					Annotations: map[string]string{"prometheus.io.scrape": "false"},
+					Annotations: annotations,
 				},
 				Spec: apiv1.PodSpec{
 					NodeSelector: nodeSelector,
@@ -267,6 +269,7 @@ func makeDeploymentSpec(request requests.CreateFunctionRequest, existingSecrets 
 }
 
 func makeServiceSpec(request requests.CreateFunctionRequest) *v1.Service {
+
 	serviceSpec := &v1.Service{
 		TypeMeta: metav1.TypeMeta{
 			Kind:       "Service",
@@ -274,7 +277,7 @@ func makeServiceSpec(request requests.CreateFunctionRequest) *v1.Service {
 		},
 		ObjectMeta: metav1.ObjectMeta{
 			Name:        request.Service,
-			Annotations: map[string]string{"prometheus.io.scrape": "false"},
+			Annotations: buildAnnotations(request),
 		},
 		Spec: v1.ServiceSpec{
 			Type: v1.ServiceTypeClusterIP,
@@ -294,6 +297,18 @@ func makeServiceSpec(request requests.CreateFunctionRequest) *v1.Service {
 		},
 	}
 	return serviceSpec
+}
+
+func buildAnnotations(request requests.CreateFunctionRequest) map[string]string {
+	var annotations map[string]string
+	if request.Annotations != nil {
+		annotations = *request.Annotations
+	} else {
+		annotations = map[string]string{}
+	}
+
+	annotations["prometheus.io.scrape"] = "false"
+	return annotations
 }
 
 func buildEnvVars(request *requests.CreateFunctionRequest) []v1.EnvVar {
--- a/handlers/update.go
+++ b/handlers/update.go
@@ -28,81 +28,123 @@ func MakeUpdateHandler(functionNamespace string, clientset *kubernetes.Clientset
 			return
 		}
 
-		getOpts := metav1.GetOptions{}
-
-		deployment, findDeployErr := clientset.ExtensionsV1beta1().
-			Deployments(functionNamespace).
-			Get(request.Service, getOpts)
-
-		if findDeployErr != nil {
-			w.WriteHeader(http.StatusNotFound)
-			w.Write([]byte(findDeployErr.Error()))
-			return
+		annotations := buildAnnotations(request)
+		if err, status := updateDeploymentSpec(functionNamespace, clientset, request, annotations); err != nil {
+			w.WriteHeader(status)
+			w.Write([]byte(err.Error()))
 		}
 
-		if len(deployment.Spec.Template.Spec.Containers) > 0 {
-			deployment.Spec.Template.Spec.Containers[0].Image = request.Image
-
-			// Disabling update support to prevent unexpected mutations of deployed functions,
-			// since imagePullPolicy is now configurable. This could be reconsidered later depending
-			// on desired behavior, but will need to be updated to take config.
-			//deployment.Spec.Template.Spec.Containers[0].ImagePullPolicy = v1.PullAlways
-
-			deployment.Spec.Template.Spec.Containers[0].Env = buildEnvVars(&request)
-
-			configureReadOnlyRootFilesystem(request, deployment)
-
-			deployment.Spec.Template.Spec.NodeSelector = createSelector(request.Constraints)
-
-			labels := map[string]string{
-				"faas_function": request.Service,
-				"uid":           fmt.Sprintf("%d", time.Now().Nanosecond()),
-			}
-
-			if request.Labels != nil {
-				if min := getMinReplicaCount(*request.Labels); min != nil {
-					deployment.Spec.Replicas = min
-				}
-
-				for k, v := range *request.Labels {
-					labels[k] = v
-				}
-			}
-
-			deployment.Labels = labels
-			deployment.Spec.Template.ObjectMeta.Labels = labels
-
-			resources, resourceErr := createResources(request)
-			if resourceErr != nil {
-				w.WriteHeader(http.StatusBadRequest)
-				w.Write([]byte(resourceErr.Error()))
-				return
-			}
-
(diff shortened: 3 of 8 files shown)


Gopkg.toml
```diff
@@ -5,7 +5,7 @@
 
 [[constraint]]
   name = "github.com/openfaas/faas"
-  version = "0.7.3"
+  version = "0.8.8"
```
ewilde: waiting on next release

alexellis: faas 0.8.8 released just now.

alexellis:
Do you see any backwards-compatibility issues?

Thanks,

Alex

alexellis:
Hi Ed,

I don't understand the 2nd test-case comment, can you clarify the meaning?

> Deployment no (but can't change label either Unexpected status: 500, message: Deployment.apps "nodejs-echo" is invalid: spec.template.metadata.labels: Invalid value)

Also this needs a rebase now due to the Gopkg changing.

Alex

ewilde:
1. @alexellis I haven't seen any backwards compatibility issues and have tried so test this as outlined in the test cases

2. I've updated the 2nd test-case comment, which hopefully clarifies the meaning

3. PR has been rebased

alexellis:
The annotation value cannot be changed at all?

alexellis:
A key part of allowing annotations is being able to update them.

It looks like this is possible with kubectl, what happened in the Kubernetes API when you tried to change an annotation value?

https://kubernetes-v1-4.github.io/docs/user-guide/kubectl/kubectl_annotate/

alexellis:
@ewilde please clarify whether annotations can be updated - i.e. 

Before:

annotation topic: payment.received

After

annotation topic: payment.cleared

Alex

ewilde:
@alexellis i've made a change locally that updates annotations successfully for deployments, replica sets using the existing api call to `DeploymentInterface.Update`. Looks like to update `Service` and `Pod` will require addition k8s api calls.

alexellis:
You shouldn't touch the Pod, Kubernetes does this automatically. I'm also unsure if we need annotations on services and deployments. Are you saying that updating a service fails if an annotation value is changed?

ewilde:
@alexellis the current implementation of the update handler before this PR uses the `DeploymentInterface.Update` k8s api to update changes to the function definiation. I've added annotations to this api call. The following k8s objects have their annotation updated now:

* `Deployment`
* `Replica set`
* `Pod`

However the service annotation is not updated by k8s automatically and retains the value from the initial deployment

ewilde:
@alexellis sorry for all the chatter on this PR :smile: 

So now we have all the resources updating their annotations
- [x] `Deployment`
- [x] `Replica set`
- [x] `Pod`
- [x] `Service`

alexellis:
This looks fine to me, after merging and testing more we'll need this in the operator too - https://github.com/openfaas-incubator/openfaas-operator

stefanprodan:
@alexellis why we need the service and replicaset to have annotations?

alexellis:
I don't think it is updating the replica set I think it's just a side effect of updating the deployment. the service seems to be annotated for consistency with how we apply the labels on the deployment. I don't know if the labels and annotations are needed on both?

stefanprodan:
I would remove the annotation and labels from the service, other than that this PR looks good.

LucasRoesler:
I thought adding these to the service was good. Removing from the replivaset makes sense to me because it is a child of the deployment, but the service doesn't have a parent to track it.

ewilde:
@LucasRoesler happy to add tests, for context there were no existing tests to `deploy.go` and `update.go` when I started working on this PR, so I was trying to fit in with the current style. However, I see your last commit added tests to `configureReadOnlyRootFilesystem` in `deploy.go`. Let me update this PR with appropriate tests

I'm not sure the `update*` methods will work as unit tests as they call the k8s api. But perhaps i'm missing something?

handlers/deploy.go
```diff
@@ -296,6 +299,18 @@ func makeServiceSpec(request requests.CreateFunctionRequest) *v1.Service {
 	return serviceSpec
 }
 
+func buildAnnotations(request requests.CreateFunctionRequest) map[string]string {
```
ewilde: @LucasRoesler @alexellis I've added unit test to this method

handlers/update.go
```diff
@@ -28,81 +28,123 @@ func MakeUpdateHandler(functionNamespace string, clientset *kubernetes.Clientset
...
+			w.WriteHeader(status)
+			w.Write([]byte(err.Error()))
 		}
+	}
+}
 
-		if len(deployment.Spec.Template.Spec.Containers) > 0 {
-			deployment.Spec.Template.Spec.Containers[0].Image = request.Image
+func updateDeploymentSpec(
+	functionNamespace string,
```
ewilde: @LucasRoesler @alexellis i've *not* added a test to this method as its not unit testable (as a side not it's not new code just extracted into a separate func from MakeUpdateHandler, to hopefully make MakeUpdateHandler easier to follow)

handlers/update.go
```diff
@@ -28,81 +28,123 @@ func MakeUpdateHandler(functionNamespace string, clientset *kubernetes.Clientset
...
+	if findServiceErr != nil {
+		return findServiceErr, http.StatusNotFound
+	}
+
+	service.Annotations = annotations
+
+	if _, updateErr := clientset.CoreV1().
+		Services(functionNamespace).
+		Update(service); updateErr != nil {
+
```
ewilde: @LucasRoesler @alexellis i've *not* added a test to this method as its not unit testable, it calls the k8s api

handlers/deploy.go
LucasRoesler: looks good

handlers/update.go
LucasRoesler: Technically it is possible to test it by using the `fake` package https://godoc.org/k8s.io/client-go/kubernetes/fake#Clientset but it probably makes sense to consider the design of the function to avoid that if we can and that can be done in the future so that we avoid changing too many things in this PR

alexellis:
I'll merge this PR before we get another rebase conflict. @stefanprodan raises a point about whether the service object should take the annotations.

PR #264, opened 2018-07-26
Enable function scheduling based on extended resource definitions.

feri:
## Description
The FunctionResources struct was extended with a new field: Others. This field stores extended resource definitions. Others is a map of strings, where the keys and values are pure strings. When the command line client (faas-cli) sends a deploy request then the deploy handler of faas-netes will process the "Others" field and populates the Limits and Requests part of the resources map respectively. These values will be then "translated" into the POD spec of the function, so that Kubernetes can schedule to nodes where those resources are available.

## Motivation and Context
The motivation behind the patch is to enable function scheduling in Kubernetes based on extended resources, such as GPUs or FPGAs. This will hopefully contribute for the fix of issue 639 in faas.

## How Has This Been Tested?
The changes -together with the patched faas-cli- were tested in a Kubernetes (v1.11.0) cluster including 4 nodes. One of the nodes is running the Intel Device Plugin and exposes the integrated Intel GPU of that computer.
During the tests I was checking if scheduling is handled properly, ie. loads (functions) requesting "intel.com/gpu" are properly directed to the node mentioned before, or they were skipped if the node "ran out" of the specified resource, or was taken away from the cluster.

## Types of changes

## Checklist:

### DIFF
--- a/handlers/deploy.go
+++ b/handlers/deploy.go
@@ -344,38 +344,62 @@ func createResources(request requests.CreateFunctionRequest) (*apiv1.ResourceReq
 		Requests: apiv1.ResourceList{},
 	}
 
-	// Set Memory limits
-	if request.Limits != nil && len(request.Limits.Memory) > 0 {
-		qty, err := resource.ParseQuantity(request.Limits.Memory)
-		if err != nil {
-			return resources, err
+	if request.Limits != nil {
+		if len(request.Limits.Memory) > 0 {
+			// Set Memory limits
+			qty, err := resource.ParseQuantity(request.Limits.Memory)
+			if err != nil {
+				return resources, err
+			}
+			resources.Limits[apiv1.ResourceMemory] = qty
 		}
-		resources.Limits[apiv1.ResourceMemory] = qty
+		if len(request.Limits.CPU) > 0 {
+			// Set CPU limits
+			qty, err := resource.ParseQuantity(request.Limits.CPU)
+			if err != nil {
+				return resources, err
+			}
+			resources.Limits[apiv1.ResourceCPU] = qty
+		}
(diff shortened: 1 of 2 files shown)


handlers/deploy.go
```diff
@@ -344,38 +344,62 @@ func createResources(request requests.CreateFunctionRequest) (*apiv1.ResourceReq
...
+			}
+			resources.Requests[apiv1.ResourceMemory] = qty
+		}
+		if len(request.Requests.CPU) > 0 {
+			// Set CPU limits
+			qty, err := resource.ParseQuantity(request.Requests.CPU)
+			if err != nil {
+				return resources, err
+			}
+			resources.Requests[apiv1.ResourceCPU] = qty
```
alexellis: This took me a while to get to. Do we need to update the deserialization code here or is it not needed?

feri: No worries. Technically the patch works as-is, considering that one uses the patched faas-cli too. However I am not totally satisfied. I believe that in the long run it would be nice to specify what exactly allowed as "extended resource". We must be able to validate and sanitize the input on the netes side. For now this sanitation happens on the client side only. There my patch makes sure that only things in the form of: vendor.domain/[gpu,fpga] are allowed to go through.

feri: The syntax for defining limits for Intel device plugins have changed as seen in this [example YAML](https://github.com/intel/intel-device-plugins-for-kubernetes/blob/master/demo/intelgpu-job.yaml#L22). 
This patch for faas-netes and the [patch for faas-cli](https://github.com/openfaas/faas-cli/pull/488) needs a re-write.

alexellis:
I'm going to close this out due to inactivity. If someone wants to revisit in the future, we'd be grateful.

PR #280, opened 2018-09-04
Add named ports to function service

alexellis:
This allows the function to be queried for metadata about which
port provides a HTTP interface.

Signed-off-by: Alex Ellis (VMware) <alexellis2@gmail.com>


alexellis:
This also pins the Dockerfiles to a newer version of Golang. In a subsequent PR we could explore Go 1.11 as a new target.

Issue #294, opened 2018-09-14
Add communication with apiserver testing when build in-cluster config


zouyee:
Add communication with apiserver testing.
## Current Behaviour

Currently， we could not check the health of client, it is a reasonable to support the health check.
## Steps to Reproduce (for bugs)

1.
2.
3.
4.

## Your Environment

* Docker version `docker version` (e.g. Docker 17.0.05 ):

* What version and distriubtion of Kubernetes are you using? `kubectl version`

* Operating System and version (e.g. Linux, Windows, MacOS):

* Link to your project or a code example to reproduce issue:

* What network driver are you using and what CIDR? i.e. Weave net / Flannel

viveksyngh:
Derek add label: proposal

alexellis:
/lock: closing as inactive. Feel free to raise a new issue if this is still required

PR #295, opened 2018-09-14
add communicating test with kubernetes apiserver

zouyee:
Signed-off-by: Zou Nengren  zouyee1989@gmail.com

## Description

Currently， we could not check the health of client, it is a reasonable to support the health check.

## Motivation and Context

- [x] I have raised an issue to propose this change ([required]
#294
(https://github.com/openfaas/faas/blob/master/CONTRIBUTING.md))

## Types of changes

## Checklist:

### DIFF
--- a/yaml/gateway-dep.yml
+++ b/yaml/gateway-dep.yml
@@ -59,6 +59,13 @@ spec:
       - name: faas-netes
         image: openfaas/faas-netes:0.6.3
         imagePullPolicy: Always
+        livenessProbe:
+          httpGet:
+            path: /healthz
+            port: 8081
+          initialDelaySeconds: 15
+          periodSeconds: 10
+          timeoutSeconds: 2
         env:
         - name: port
           value: "8081"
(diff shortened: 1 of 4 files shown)


alexellis:
Do we need this? If so, shouldn't it be part of a long-running go-routine that indicates the status via /healthz?

yaml/gateway-dep.yml
```diff
@@ -59,6 +59,13 @@ spec:
       - name: faas-netes
         image: openfaas/faas-netes:0.6.0
         imagePullPolicy: Always
+        livenessProbe:
```
alexellis: This would have to be added to the helm chart also.

stefanprodan:
@alexellis why we need this check in the livenessProbe? A controller should survive a short k8s API downtime without crash looping. With this change when you do control plane upgrade faas-netes will crash.

yaml/gateway-dep.yml
zouyee: done

alexellis:
@stefanprodan I'm not sure why you're asking me that question? I suggested that we don't need this PR at all. What do we want to do?

PR #317, opened 2018-11-07
Add external svc for prometheus helm chart

### LINKED ISSUE #316: Prometheus is not by default NodePorts as API Gateway in helm.
In README.md it said that:
`By default NodePorts will be created for the API Gateway and Prometheus.` which is actually not true. Prometheus is now by default ClusterIP, should we make it NodePorts?

LucasRoesler:
## Description
- Add external srv to expose prometheus as a node port 31119 so that the
installation matches the README description.  This also matches the static
yaml manifest installs

## Motivation and Context

Resolves #316

## How Has This Been Tested?

Tested by installing on a local Docker-for-Mac and seeing the new `prometheus-external` svc is created

## Types of changes

## Checklist:

### DIFF
--- a/chart/openfaas/templates/prometheus-external-svc.yaml
+++ b/chart/openfaas/templates/prometheus-external-svc.yaml
@@ -1,0 +1,24 @@
+{{- if .Values.exposeServices }}
+---
+apiVersion: v1
+kind: Service
+metadata:
+  labels:
+    app: {{ template "openfaas.name" . }}
+    chart: {{ .Chart.Name }}-{{ .Chart.Version }}
+    component: prometheus
+    heritage: {{ .Release.Service }}
+    release: {{ .Release.Name }}
+  name: prometheus-external
+  namespace: {{ .Release.Namespace | quote }}
+spec:
+  type: {{ .Values.serviceType }}
+  ports:
+    - port: 9090
+      protocol: TCP
+      {{- if contains "NodePort" .Values.serviceType }}
+      nodePort: {{ .Values.prometheus.nodePort }}
+      {{- end }}
+  selector:
+    app: prometheus
+{{- end }}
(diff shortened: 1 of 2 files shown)


alexellis:
This was removed under guidance from @stefanprodan. We wanted to keep the surface area low for helm and left it in the YAML for development. 

Perhaps this should be a flag in the chart or we should document the port-forward command more prominently?

Interested in thoughts from the two of you.

stefanprodan:
We could have an option in Helm to expose Prometheus but I think it should be off by default. Prometheus has no auth and I think by we should not expose anything else beside the Gateway, Prom is an implementation detail of OF not a UI that users should use. Maybe a better alternative would be to add Grafana to the helm chart since it does support auth.

LucasRoesler:
Then we should also remove the same node port on the yaml manifests, right?

alexellis:
No, see this comment:

https://github.com/openfaas/faas-netes/pull/317#issuecomment-436574117

The decision was made to keep the NodePort in the YAML files which you cannot customize, these are not production-ready for every use-case or environment. After this was done I updated the documentation to make it clearer:

https://docs.openfaas.com/deployment/kubernetes/#pick-helm-or-yaml-files-for-deployment-a-or-b

> It is recommended to use helm to install OpenFaaS so that you can configure your installation to suit your needs. This configuration is considered to be production-ready.

> Plain YAML files are also provided for x86_64 and armhf, but since they cannot be customized easily it is recommended that you only use these for local development.

alexellis:
I think we might need to take some additional actions, but I don't know which at this time. Perhaps the README should be trimmed a little and point to the docs?

LucasRoesler:
I made a similar comment in the original issue, if we are good with that, I have a branch ready.

LucasRoesler:
@alexellis for now, I created a PR to update the README see #319

PR #323, opened 2018-11-15
Simplify handlers

wpjunior:
Two smalls tips to simplify handlers
- Remove Body.Close() call from handlers
- Use json.Decoder instead of JSON.Unmarshall

### DIFF
--- a/handlers/delete.go
+++ b/handlers/delete.go
@@ -17,8 +17,6 @@ import (
 // MakeDeleteHandler delete a function
 func MakeDeleteHandler(functionNamespace string, clientset *kubernetes.Clientset) http.HandlerFunc {
 	return func(w http.ResponseWriter, r *http.Request) {
-		defer r.Body.Close()
-
 		request := requests.DeleteFunctionRequest{}
 		err := json.NewDecoder(r.Body).Decode(&request)
 		if err != nil {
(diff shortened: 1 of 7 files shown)


alexellis:
Derek close

alexellis:
Hi @wpjunior this PR would change the behaviour of the handlers in ways we don't want. In our experience the body must always be closed of the request to prevent the connection from being kept open.

We also have an established standard for using `Unmarshal/Marshal` so don't want to introduce inconsistencies.

If you'd like to get involved with the project and contribute then please let me know and I'll invite you to the Slack community and we can try to find you something to get started with?

Alex

wpjunior:
Hi @alexellis, Thanks for your invitation to the slack project, I can contribute with ideas and discussion to improve this project, could you send me an invitation later, no problem =)

I understood the project standard to use `json.Unmarshal/Marshal`, IMHO this rule is very important to the project quality, I prefer `json.Encoder` by the low number of variables and lines required to decode any stream into Go-Like structure/interface{}, but standards are standards =)

And the drop of `(net/http. *Request).Body.Close()` in server handle-side, it's so required just in client-side applications, https://github.com/golang/go/blob/master/src/net/http/request.go#L174, But I think it's would be necessary to developers use the same-way in server/client, I prefer to `Close()` body of the request just in client-side applications, to preserve the consistency to the standard library documentation.

Thanks =)

Issue #354, opened 2019-01-15
Multiple gateways are not detected by Prometheus


alexellis:
## Expected Behaviour

If I have multiple gateway replicas in Kubernetes, I expect metrics to be scraped from each of them. This not the case with our current discovery based upon DNS-A records.

## Current Behaviour

The metrics are inconsistent and appear not to update.

## Possible Solution

Via Stefan we could use Pod discovery:

https://github.com/stefanprodan/k8s-prom-hpa/blob/master/prometheus/prometheus-cfg.yaml#L202

## Steps to Reproduce (for bugs)

1. Scale the gateway to > 1 replica and invoke a function
2. Use `kubectl port-forward` to bring back Prometheus to your local machine to look at the targets page for the amount of gateways being scraped.

## Context

This probably doesn't affect most users who I believe run only one gateway replica, however we recommend > 1 for HA.

Docker Swarm is unaffected.

## Your Environment

Present in the current version of helm / YAML config.

alexellis:
This is a community project so the community is free to take responsibility for "movement" on things :smile:

As far as I can see nobody has volunteered to work on the fix. If you'd like you can try to do it and use the link provided by Stefan? You will also need to setup some RBAC permissions to allow Prometheus to list Pods.

martindekov:
Hey Alex, I have set up Kubernetes Cluster on Digital Ocean for this one and it is nice to have OF instance in general somewhere, I won't be able to actively work on the Issue this and the next week since university is calling(exams). Just wanted to let you now in case this is critical issue.

alexellis:
Let's see if anyone else can pick it up in #kubernetes?

martindekov:
Ok I will ask

PR #363, opened 2019-01-27
 Use Kubernetes discovery for Gateway Prometheus scrapping 

stefanprodan:
This PR does the following:
- upgrade Prometheus to v2.6.1
- add Prometheus Kubernetes service account, role and RBAC
- enable pod scraping for OpenFaaS Gateway on port 8080
- disable Istio sidecar injection except for Gateway and QueueWorker

Fix: #354

## Motivation and Context

## How Has This Been Tested?
Helm install tested on GKE with Istio 1.0.3 and two Gateway pods

Prometheus:

screenshot 2019-01-27 at 12 02 22

Gateway promql:

screenshot 2019-02-03 at 12 19 37

Grafana:

screenshot 2019-02-03 at 12 20 38

## Types of changes

## Checklist:

### DIFF
--- a/chart/openfaas/templates/alertmanager-dep.yaml
+++ b/chart/openfaas/templates/alertmanager-dep.yaml
@@ -16,6 +16,7 @@ spec:
       labels:
         app: alertmanager
       annotations:
+        sidecar.istio.io/inject: "false"
         checksum/alertmanager-config: {{ include (print $.Template.BasePath "/alertmanager-cfg.yaml") . | sha256sum | quote  }}
     spec:
       containers:
--- a/chart/openfaas/templates/faas-idler-dep.yaml
+++ b/chart/openfaas/templates/faas-idler-dep.yaml
@@ -10,6 +10,7 @@ spec:
   template:
     metadata:
       annotations:
+        sidecar.istio.io/inject: "false"
         prometheus.io.scrape: "false"
       labels:
         app: faas-idler
--- a/chart/openfaas/templates/faas-idler-dep.yaml
+++ b/chart/openfaas/templates/faas-idler-dep.yaml
@@ -10,7 +10,6 @@ spec:
   template:
     metadata:
       annotations:
-        sidecar.istio.io/inject: "false"
         prometheus.io.scrape: "false"
       labels:
         app: faas-idler
--- a/chart/openfaas/templates/prometheus-cfg.yaml
+++ b/chart/openfaas/templates/prometheus-cfg.yaml
@@ -26,13 +26,31 @@ data:
         static_configs:
           - targets: ['localhost:9090']
 
-      - job_name: "gateway"
+      - job_name: 'kubernetes-pods'
         scrape_interval: 5s
-        dns_sd_configs:
-          - names: ['gateway.{{ .Release.Namespace }}']
-            port: 8080
-            type: A
-            refresh_interval: 5s
+        honor_labels: false
+        kubernetes_sd_configs:
+          - role: pod
+            namespaces:
+              names:
+                - {{ .Release.Namespace }}
+        relabel_configs:
+        - action: labelmap
+          regex: __meta_kubernetes_pod_label_(.+)
+        - source_labels: [__meta_kubernetes_namespace]
+          action: replace
+          target_label: kubernetes_namespace
+        - source_labels: [__meta_kubernetes_pod_name]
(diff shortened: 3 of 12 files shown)


alexellis:
> disable Istio sidecar injection except for Gateway and QueueWorker

Should it not be on for Prometheus and AlertManager?

alexellis:
What about faas-idler?

alexellis:
Please can you confirm if the name or labels of metrics has changed by switching to this mode of discovery?

Maybe showing the Grafana dashboard from the labs would be sufficient?

https://github.com/openfaas/workshop/blob/kubernetes-preview/lab2.md

alexellis:
@stefanprodan can you raise an issue to track the problems you had with Istio? I don't think those changes belong in this PR, but it might be because I don't have all the context.

alexellis:
Hi Stefan thank you for updating and testing the newer version. Please could you look at my comments about the istio changes? ideally I would prefer to take these in a separate PR and have an issue explaining more about the context.

Thanks again for your contribution

Alex

stefanprodan:
I've added an issue for Istio see #369

chart/openfaas/templates/alertmanager-dep.yaml
```diff
@@ -16,6 +16,7 @@ spec:
       labels:
         app: alertmanager
       annotations:
+        sidecar.istio.io/inject: "false"
```
alexellis: Just curious.. why doesn't this work for AlertManager? Because of its HA discovery mechanism?

chart/openfaas/templates/faas-idler-dep.yaml
```diff
@@ -10,6 +10,7 @@ spec:
   template:
     metadata:
       annotations:
+        sidecar.istio.io/inject: "false"
```
alexellis: Why?

chart/openfaas/templates/prometheus-cfg.yaml
```diff
@@ -26,13 +26,31 @@ data:
...
-      - job_name: "gateway"
+      - job_name: 'kubernetes-pods'
         scrape_interval: 5s
-        dns_sd_configs:
-          - names: ['gateway.{{ .Release.Namespace }}']
-            port: 8080
-            type: A
-            refresh_interval: 5s
+        honor_labels: false
+        kubernetes_sd_configs:
```
alexellis: Thanks this looks pretty complicated compared to before. Do you have a source of where this comes from?

chart/openfaas/templates/prometheus-dep.yaml
```diff
@@ -16,8 +16,10 @@ spec:
       labels:
         app: prometheus
       annotations:
+        sidecar.istio.io/inject: "false"
```
alexellis: How come?

chart/openfaas/values.yaml
```diff
@@ -51,7 +51,7 @@ operator:
 
 # monitoring and auto-scaling components
 prometheus:
-  image: prom/prometheus:v2.3.1
+  image: prom/prometheus:v2.7.1
```
alexellis: @rdimitrov looks like we should use `v2.7.1` for Docker Swarm, we used `2.7.0` which may be my fault.

chart/openfaas/templates/prometheus-dep.yaml
stefanprodan: The K8s SD returns pod IPs and those don't work inside the mesh

chart/openfaas/templates/prometheus-cfg.yaml
stefanprodan: Here it is https://github.com/prometheus/prometheus/blob/master/documentation/examples/prometheus-kubernetes.yml

chart/openfaas/templates/prometheus-dep.yaml
stefanprodan: There is a way to make it work but it needs the Istio cert, here is an example https://github.com/stefanprodan/flagger/blob/master/artifacts/gke/istio-prometheus.yaml#L330

alexellis:
Merged, thank you Stefan.

alexellis:
Tested on GKE with 4x GWs and `hey` - stats reported correctly 👍

alexellis:
Derek lock

PR #373, opened 2019-02-17
Fix remove value from scaling alert definition

### LINKED ISSUE #372: Potential issues with AlertManager
## Expected Behaviour

Given a stable load from `hey` such as `-c 2 -q 10 -n 100000` into the `echo-go` function, I would expect to see a steady rate of increase on gateway invocations and in the service count.

## Current Behaviour

* Sometimes AlertManager continues to fire alerts, even after the Alert is cleared in the Alerts tab in Prometheus
* Sometimes the Alert appears to be stuck as "Pending"
* Sometimes the resolved alert is taking much longer than usual to fire
* Sometimes the Alert has hysteresis and is firing/resolving continually instead of steadily increasing the replica count. 

I believe this was introduced between 1.8.1 and 1.9.1 of the helm chart.

screenshot 2019-02-16 at 11 13 12

Kubeadm / 1.13 single node

screenshot 2019-02-16 at 11 21 07

Swarm

## Investigate

Test each version with `kind` or similar using a script for a repeatable environment, until we can figure out what changed / went wrong.

Compare to Swarm

Changes and variables:

* Prometheus version upgraded https://github.com/openfaas/faas-netes/commit/37108946eb8352c200eed2c893ab2041e76cba43
* Basic auth for /system/alert added and AlertManager configs updated for this https://github.com/openfaas/faas-netes/pull/357
* Kubernetes SD enabled for Pod scraping by Prometheus https://github.com/openfaas/faas-netes/pull/363

Not changed:

* AlertManager *not* upgraded, 0.16.1 is available with many changes since 0.15.0 that we are using now.

Prom queries:

```
sum by(function_name) (rate(gateway_function_invocation_total{code="200"}[10s]))

gateway_service_count

http_requests_total{path="/system/alert"}
```

Loading:

```
git clone https://github.com/alexellis/echo-fn
faas-cli deploy -g ...

hey -c 2 -q 10 -n 10000 http://127.0.0.1:8080/function/go-echo

```

Port-forwarding:

```
kubectl port-forward svc/prometheus -n openfaas 9090:9090
```

LucasRoesler:
Updates the alert configuration and updates alert manager to the latest version.

## Description

- upgrade alert manager to the latest release to match prometheus and to
get the expected alert behavior
- Removes the `alert` label in the scale-up alert
- Updates the annotaitons to use the `function_name` label instead of
the `instance` label that was removed.
- Per prometheus/prometheus#4836 and the related mailing list discussion
https://groups.google.com/d/msg/prometheus-users/7Ul6ngc7Ogs/j_YDszV5BwAJ
the alert value should not be included in the alert labels otherwise
each calculation of the alert is treated like a new alert and then the
use of `for 5s` will not behave as expected.

## Motivation and Context

Resolves #372 

## How Has This Been Tested?

Manually tested locally, 

1. deploy the helm chart 
```sh
helm upgrade openfaas --install ./openfaas \
    --namespace openfaas  \
    --set basic_auth=false \
    --set functionNamespace=openfaas-fn
```
2. deploy the go-echo function https://github.com/alexellis/echo-fn
3. create constant load using `hey`
```sh
hey -c 2 -q 10 -n 100000 http://127.0.0.1:31112/function/go-echo
```
4. observe that the function scales in steps relatively quickly after the alert stops and that the function then scales quickly to 1 replica when the alert stops, this is shown in the below graphs

image

Note: you should be able to reproduce those graphs by 
1. forwarding the prometheus pod to 9090
2. opening  http://localhost:9090/graph?g0.range_input=15m&g0.expr=rate(gateway_function_invocation_total%5B10s%5D)&g0.tab=0&g1.range_input=15m&g1.expr=ALERTS%7Balertname%3D%22APIHighInvocationRate%22%7D&g1.tab=0&g2.range_input=15m&g2.expr=gateway_service_count&g2.tab=0

## Types of changes

## Checklist:

### DIFF
--- a/chart/openfaas/templates/prometheus-cfg.yaml
+++ b/chart/openfaas/templates/prometheus-cfg.yaml
@@ -70,7 +70,6 @@ data:
           labels:
             service: gateway
             severity: major
-            value: '{{ "{{" }}$value{{ "}}" }}'
           annotations:
-            description: High invocation total on "{{ "{{" }}$labels.instance{{ "}}" }}"
-            summary: High invocation total on "{{ "{{" }}$labels.instance{{ "}}" }}"
+            description: High invocation total on "{{ "{{" }}$labels.function_name{{ "}}" }}"
+            summary: High invocation total on "{{ "{{" }}$labels.function_name{{ "}}" }}"
--- a/chart/openfaas/values.yaml
+++ b/chart/openfaas/values.yaml
@@ -55,7 +55,7 @@ operator:
 prometheus:
   image: prom/prometheus:v2.7.1
 alertmanager:
-  image: prom/alertmanager:v0.15.0
+  image: prom/alertmanager:v0.16.1
 
 # async provider
 nats:


alexellis:
Thanks for putting this together.

I think that we need to test with `basic_auth` on and off, at least with it on as that is the default as per the docs and expected production usage.

Alex

LucasRoesler:
I just retested with basic auth enabled

```sh
helm upgrade openfaas --install ./openfaas \
    --namespace openfaas  \
    --set basic_auth=true \
    --set functionNamespace=openfaas-fn
```

I confirmed that the UI and CLI required the basic auth and then tested using the same `hey` command

```sh
hey -c 2 -q 10 -n 100000 http://127.0.0.1:31112/function/go-echo
```

And I am seeing the same, correct, behavior 

image

That graph uses a 5 min window instead of a 15 min window, [link for 5min window](http://localhost:9090/graph?g0.range_input=5m&g0.expr=rate(gateway_function_invocation_total%5B10s%5D)&g0.tab=0&g1.range_input=5m&g1.expr=ALERTS%7Balertname%3D%22APIHighInvocationRate%22%7D&g1.tab=0&g2.range_input=5m&g2.expr=gateway_service_count&g2.tab=0)

As a side note, I think that future changes that we are concerned impact scaling should include a screenshot of the above graphs as part of the review _before_ we merge.  I feel pretty comfortable that reproducing those graphs indicates healthy scaling behavior.

alexellis:
This looks much better, would you like me to merge since it resolves the problems we've seen so far?

I think we'd need similar for Swarm, too?

Whilst you have everything set up, I have two other test suggestions:

* a longer test with the `-z` parameter which specifies a time duration rather than amount of requests
* multiple test runs returning to zero replicas each time (to see if there is any effect of trailing alerts)

Alex

LucasRoesler:
Here is a test with scale to zero

```sh
helm upgrade openfaas --install ./openfaas \
    --namespace openfaas  \
    --set basic_auth=true \
    --set functionNamespace=openfaas-fn \
    --set faasIdler.dryRun=false \
    --set faasIdler.inactivityDuration=1m
```

image

I  have also tested with multiple runs scaling up and down,then to zero, then up again, back to 1, then up, then down to 1, then 0, then up again.  Basically multiple repeated spread out bursts of load seem to behave exactly as you would expected

LucasRoesler:
Should this also bump the chart version or would you do that as a separate PR?

alexellis:
Follow-up with another PR please for the bump and publish.

You have to run `make charts` in the root too.

alexellis:
Merged 🎉 thank you everyone

alexellis:
Derek lock

Issue #352, opened 2019-01-11
Test helm chart in CI using KinD


alexellis:
## Expected Behaviour

We can test the helm chart and yaml files during builds by using the KinD tool in the Travis build it only needs Docker and kubectl to be installed. 

## Current Behaviour

Manual testing 

## Possible Solution

We should also run the certifier tool / tests as part of this work. 

Example:  

https://blog.alexellis.io/get-started-with-openfaas-and-kind/

https://github.com/openfaas/certifier

alexellis:
Booting a KinD cluster took about 90s - adding 1m30s.

zeerorg:
Are you working with this issue? Or can I take this up ?

burtonr:
@zeerorg the `help wanted` tag indicates that it's up for grabs (as long as nobody else has mentioned they would take it). Let us know if you have any questions or need any help!

zeerorg:
I think alex has done some work in this PR https://github.com/openfaas/faas-netes/pull/375

burtonr:
hmm... looks like you're right. I guess this will have to wait until @alexellis can respond. I didn't see that change as there is no indication of that in the PR, only in the commit messages (and files changed)

good catch!

alexellis:
I have started on this, but have deliberately left the testing stage unimplemented for the community to pick up.

zeerorg:
I'll pick this issue up. I've been wanting to get more into e2e testing with kubernetes and this seems like a good starting point. 

We can also look at k3s for testing since it has lesser starting time, though it can be done later on to improve performance

zeerorg:
Derek assign: me

alexellis:
Derek close

alexellis:
@zeerorg I was just chatting with @LucasRoesler and noticed that the certifier hadn't been done as part of #382.

I'd be happy for either of you to pick that up. I'll re-open the issue.

zeerorg:
Ceritifer skipped my mind. I'll add that in a PR.

zeerorg:
Derek assign: me

alexellis:
@LucasRoesler would you like to take over please with a fresh PR?

alexellis:
Basically, I would think a copy/paste of your instructions in the README of the certifier wouldn't be far off the mark.

LucasRoesler:
Derek assign: me

alexellis:
/lock: closing as inactive. Feel free to raise a new issue if this is still required

PR #382, opened 2019-03-07
CI for Helm using KinD

zeerorg:
Signed-off-by: Rishabh Gupta <r.g.gupta@outlook.com>

## Description
Changes:
- Removed tiller deploy for CI
- Deploy helm chart using helm templates
- Install faas-cli and create a function "echo"
- Call the function and check if the response is OK

## Motivation and Context
Fixes Issue: https://github.com/openfaas/faas-netes/issues/352

## Types of changes
Adds testing

## Checklist:


.travis.yml
```diff
@@ -24,7 +24,8 @@ script:
   - ./contrib/get_kind.sh
   - ./contrib/create_kind_cluster.sh
   - ./contrib/deploy.sh
-  - ./contrib/install_tiller.sh
```
alexellis: Where did install_tiller.sh go?

stefanprodan:
Hi @zeerorg 

I would use Tiller instead of helm template, here are the step for that:

```bash
echo ">>> Installing Helm"
curl https://raw.githubusercontent.com/kubernetes/helm/master/scripts/get | bash

echo '>>> Installing Tiller'
kubectl --namespace kube-system create sa tiller
kubectl create clusterrolebinding tiller-cluster-rule --clusterrole=cluster-admin --serviceaccount=kube-system:tiller
helm init --service-account tiller --upgrade --wait
```

contrib/run_function.sh
```diff
@@ -0,0 +1,37 @@
...
+
+export KUBECONFIG="$(kind get kubeconfig-path)"
+
+kubectl rollout status deploy/gateway -n openfaas
+kubectl port-forward deploy/gateway -n openfaas 8080:8080 &
+
+# Check if Openfaas GW is Up
+for i in {1..60};
+do
+    curl -if 127.0.0.1:8080
```
alexellis: Or `/healthz`?

contrib/run_function.sh
```diff
@@ -0,0 +1,37 @@
+#!/bin/bash
+
+export KUBECONFIG="$(kind get kubeconfig-path)"
+
+kubectl rollout status deploy/gateway -n openfaas
+kubectl port-forward deploy/gateway -n openfaas 8080:8080 &
+
+# Check if Openfaas GW is Up
+for i in {1..60};
```
alexellis: Unlike to be enough?

alexellis:
@stefanprodan Thanks for reviewing. We actually had tiller here, but it looks like it got removed for some reason @zeerorg?

contrib/run_function.sh
stefanprodan: I think this doesn't need to spin wait since `kubectl rollout status` includes the health check.

contrib/run_function.sh
```diff
@@ -0,0 +1,37 @@
...
+
+# Login in OpenFaas
+export OPENFAAS_URL=http://127.0.0.1:8080
+PASSWORD="something_random" # Set to predetermined value from ./contrib/deploy.sh
+faas-cli login --username admin --password $PASSWORD
+
+faas-cli deploy --image=functions/alpine:latest --fprocess=cat --name echo
+
+# Call echo function
+for i in {1..60};
```
stefanprodan: I would make this 180

zeerorg: I'll update this 👍

.travis.yml
zeerorg: I am still learning a bit of kubernetes, I've pushed update to install tiller. My reasoning was that it might not add much value to the test and I could remove it to save time on testing, though I've added it now.

contrib/deploy.sh
```diff
@@ -1,3 +1,19 @@
...
+
+echo "Applying namespaces"
+kubectl apply -f ./namespaces.yml
+
+PASSWORD="something_random"
+kubectl -n openfaas create secret generic basic-auth \
+--from-literal=basic-auth-user=admin \
+--from-literal=basic-auth-password="$PASSWORD"
+
+helm install ./chart/openfaas \
```
stefanprodan: This is not how install works, you don't need to generate a file or apply it with kubectl. Please use `install --wait`

zeerorg:
I'll incorporate all changes and push the PR once again. Thanks for the reviews @stefanprodan @alexellis I'm still learning about helm and tiller

contrib/deploy.sh
zeerorg: Did `install --wait`

contrib/run_function.sh
zeerorg: I has removed it in one of my commit, but found out that the test gave some errors about gateway not being up, not really sure why it happened, but seems like port-forward takes some time to warm up. See https://travis-ci.org/openfaas/faas-netes/builds/502995868#L789 from latest build

zeerorg: Instead of Checking for up, I've added a sleep step

LucasRoesler:
looks good to me 🚀

contrib/install_tiller.sh
```diff
@@ -7,15 +7,4 @@ kubectl -n kube-system create sa tiller \
       --clusterrole cluster-admin \
       --serviceaccount=kube-system:tiller
 
-helm init --skip-refresh --upgrade --service-account tiller
```
alexellis: Why aren't we checking for tiller to be ready?

contrib/run_function.sh
```diff
@@ -0,0 +1,28 @@
+#!/bin/bash
+
+export KUBECONFIG="$(kind get kubeconfig-path)"
+
+kubectl rollout status deploy/gateway -n openfaas
```
alexellis: What does this line do?

LucasRoesler: that should wait for the gateway deployment to register as healthy, i just discovered this command recently as well from my ops team

alexellis: Is it blocking? What is the timeout?

LucasRoesler: Yes it blocks, it returns a nonzero exit code if the deployment misses the progression deadline. I am not sure what that value is, but it is finite

alexellis: We should find out and pass it as a flag if available to self document

contrib/install_tiller.sh
zeerorg: I added a `--wait` flag at the end of the script

contrib/run_function.sh
zeerorg: Ohh. good catch, I should add a check to see if rollout succeeds

contrib/install_tiller.sh
alexellis: OK that resolves my concern. Can you back-port these changes to off-bootstrap too? Looks 💯

contrib/run_function.sh
alexellis:
```
      --timeout=0s: The length of time to wait before ending watch, zero means never. Any other values should contain a
corresponding time unit (e.g. 1s, 2m, 3h).
```

I'd like to see explicit waits done.

contrib/deploy.sh
```diff
@@ -1,3 +1,18 @@
 #!/bin/bash
 
 export KUBECONFIG="$(kind get kubeconfig-path)"
+
+echo "Applying namespaces"
+kubectl apply -f ./namespaces.yml
+
+PASSWORD="something_random"
```
alexellis: Hmm? What about using something random and storing it in a file?

contrib/get_faas.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+
+sudo curl -sL https://cli.openfaas.com | sudo sh
```
alexellis: why do `sudo curl`?

contrib/run_function.sh
```diff
@@ -0,0 +1,34 @@
...
+sleep 10
+
+# Login in OpenFaas
+export OPENFAAS_URL=http://127.0.0.1:8080
+PASSWORD="something_random" # Set to predetermined value from ./contrib/deploy.sh
+faas-cli login --username admin --password $PASSWORD
+
+faas-cli deploy --image=functions/alpine:latest --fprocess=cat --name echo
+
+# Call echo function
```
alexellis: We could check readiness with `faas-cli describe` or the API?

contrib/get_faas.sh
zeerorg: Should I do it without sudo ? travis doesn't run scripts in root by default so sudo was needed for that

contrib/deploy.sh
zeerorg: I wanted to make the test easy to follow. Saving it a file seemed like a bit complex 🤔  it can be done later if there is any issue. I am also using the password to login in faas-cli in `run_function.sh`

alexellis: Please store it in a file and use it again?

contrib/get_faas.sh
alexellis: You only need `sudo` for the `sh` not for `curl` which doesn't require privileges.

contrib/run_function.sh
```diff
@@ -0,0 +1,35 @@
+#!/bin/bash
+
+export KUBECONFIG="$(kind get kubeconfig-path)"
+
+kubectl rollout status deploy/gateway -n openfaas --timeout=0s
```
alexellis: This is not what I asked for. I assume timeout=0s means infinite? We definitely need at timeout, set it to something sensible and we can tune it over time.

zeerorg: Added "1m" as timout.

contrib/run_function.sh
```diff
@@ -0,0 +1,36 @@
+#!/bin/bash
+
+export KUBECONFIG="$(kind get kubeconfig-path)"
+
+kubectl rollout status deploy/gateway -n openfaas --timeout=1m
```
alexellis: `1m` seems a bit hopeful, but I'll merge.

alexellis:
Thanks for working on this.

zeerorg:
Anytime 😄

PR #401, opened 2019-03-31
Parameterize connector deployment name

### LINKED ISSUE #400: kafka-connector helm chart - make deployment name a parameter
## Expected Behaviour

I should be able to install the kafka-connector helm chart twice into the same namespace with varying names such as:

```
kafka-connector-payments
kafka-connector-data-imports
```

## Current Behaviour

Can only be installed into the same namespace once

## Possible Solution

Allow the Deployment name to be overridden or suffixed 

See instructions for installing chart: https://github.com/openfaas/faas-netes/tree/master/chart/kafka-connector

LucasRoesler:
## Description

- use the fullname template helper instead of a constant.  This allows
the chart to be installed multiply times and the user can override the
value via the "fullnameOverride"

## Motivation and Context

Resolves #400

## How Has This Been Tested?

Observed the deployment name value changes to match the override when using `--dry-run` and `--debug` to deploy it.

## Types of changes

## Checklist:

### DIFF
--- a/chart/kafka-connector/README.md
+++ b/chart/kafka-connector/README.md
@@ -44,15 +44,15 @@ TBD
 
 Additional kafka-connector options in `values.yaml`.
 
-| Parameter          | Description                                                                            | Default                        |
-| ------------------ | -------------------------------------------------------------------------------------- | ------------------------------ |
-| `upstream_timeout` | Maximum timeout for upstream function call, must be a Go formatted duration string.    | `30s`                          |
-| `rebuild_interval` | Interval for rebuilding function to topic map, must be a Go formatted duration string. | `3s`                           |
-| `topics`           | Topics to which the connector will bind, provide as a comma-separated list.            | `faas-request`                 |
-| `gateway_url`      | The URL for the API gateway.                                                           | `http://gateway.openfaas:8080` |
-| `broker_host`      | location of the Kafka brokers.                                                         | `kafka`                        |
-| `print_response`   | Output the response of calling a function in the logs.                                 | `true`                         |
-| `fullnameOverride` | Override the name value used for the Connector Deployment object.                      | ``                             |
+| Parameter                | Description                                                                            | Default                        |
+| ------------------------ | -------------------------------------------------------------------------------------- | ------------------------------ |
+| `upstream_timeout`       | Maximum timeout for upstream function call, must be a Go formatted duration string.    | `30s`                          |
+| `rebuild_interval`       | Interval for rebuilding function to topic map, must be a Go formatted duration string. | `3s`                           |
+| `topics`                 | Topics to which the connector will bind, provide as a comma-separated list.            | `faas-request`                 |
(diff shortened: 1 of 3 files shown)


chart/kafka-connector/README.md
```diff
@@ -52,6 +52,7 @@ Additional kafka-connector options in `values.yaml`.
 | `gateway_url`      | The URL for the API gateway.                                                           | `http://gateway.openfaas:8080` |
 | `broker_host`      | location of the Kafka brokers.                                                         | `kafka`                        |
 | `print_response`   | Output the response of calling a function in the logs.                                 | `true`                         |
+| `fullnameOverride` | Override the name value used for the Connector Deployment object.                      | ``                             |
```
alexellis: I think this was meant to match the template which uses `connector.fullname`?

alexellis: Can we we use something else like `deploymentName`?

chart/kafka-connector/templates/deployment.yml
```diff
@@ -16,7 +16,7 @@ metadata:
     app.kubernetes.io/part-of: openfaas
     app.kubernetes.io/managed-by: {{ .Release.Service }}
     helm.sh/chart: {{ .Chart.Name }}-{{ .Chart.Version | replace "+" "_" }}
-  name: kafka-connector
+  name: {{ template "connector.fullname" . }}
```
alexellis: What is this going to output if nobody defines `fullname`?

LucasRoesler: This is using the typical full name helper that many charts have 


```
{{- define "connector.fullname" -}}
{{- if .Values.fullnameOverride -}}
{{- .Values.fullnameOverride | trunc 63 | trimSuffix "-" -}}
{{- else -}}
{{- $name := default .Chart.Name .Values.nameOverride -}}
{{- if contains $name .Release.Name -}}
{{- .Release.Name | trunc 63 | trimSuffix "-" -}}
{{- else -}}
{{- printf "%s-%s" .Release.Name $name | trunc 63 | trimSuffix "-" -}}
{{- end -}}
{{- end -}}
{{- end -}}
```

If `fullnameOverride` is set, it completely overrides the deployment name
Otherwise, the name is either _nameOverride_ if it is defined or default to _the chart name_ and then the name is constructed at `releaseName-name` truncated to 63 characters.

chart/kafka-connector/README.md
LucasRoesler: sure

alexellis: If `fullnameOverride` is the standard, then this is my misunderstanding and we should revert to what you had.

chart/kafka-connector/templates/deployment.yml
alexellis: So we don't need to add the macro above, it's just automatic?

alexellis:
Thanks for your PR, I'll revert the change myself to your original value of `fullnameOverride`. I didn't know from the PR / context that `fullnameOverride` is a convention.

Issue #349, opened 2019-01-10
Create instructions for faas-netes on OpenShift


alexellis:
We should create instructions for deploying OpenFaaS on OpenShift.

Michael Hausenblas once worked on this and wrote up a blog using `minishift`, but it wasn't signed-off, so couldn't be merged. The project has changed since then and I'd appreciate an experienced OpenShift user taking a look at this and contributing changes back to this repo and openfaas/docs.

annatech:
Having recently deployed faas-netes on a self-hosted OpenShift OKD 3.11 cluster, I can follow-up with a document describing my work. Overall, I was able to use the instructions from here, mostly as-is:

https://github.com/openfaas/faas-netes/blob/master/chart/openfaas/README.md

The actual deployment was performed using a helm chart, executed from one of the OpenShift master nodes. After the kubectl command was run to create the namespaces and helm deployed the containers, I moved to OpenShift for the rest of the configuration. Since I had already setup a fully-functional and ssl-secured cluster, service routes were easy to deploy over HTTPS. 

Finally, I did have to create a network bridge between the openfaas and openfaas Projects (a.k.a "namespaces" in k8s). This was primarily due to my having chosen the multi-tenant network plugin which segments Pods between projects by default.

The advantage of deploying on OpenShift is that container management and route configuration is a breeze. Tasks which are a chore in pure k8s are often entirely UI-driven or orchestrated by compact "oc" CLI commands.  The RBAC, network segmentation and cluster management tools make it better suited for "enterprise" deployments. The downside is that the minimum requirements are much higher than, say, Docker Swarm... so no Raspberry Pi's for this one.

alexellis:
Please send a PR to appear here https://docs.openfaas.com/deployment/openshift/

Through editing this file:

https://github.com/openfaas/docs/blob/master/docs/deployment/openshift.md

It should cover:

* minishift installation/setup
* OC on-prem setup

Using helm + tiller or helm to generate YAML for `kubectl apply -f`

alexellis:
the latest update -> https://blog.alexellis.io/openshift-in-a-footloose-container/

PR #407, opened 2019-04-08
Add in routes for OpenShift

mirroredge:
Add in OpenShift routes to make it compatible with OpenShift 3.11
## Description
Created two routes, one for the gateway and prometheus so they are externally exposed.

## Motivation and Context
https://github.com/openfaas/faas-netes/issues/349

## How Has This Been Tested?
By running the deployment instructions here:
https://github.com/mirroredge/docs
Soon to be merged into the openfaas docs repo

## Types of changes

## Checklist:

PR incoming for documentation


openshift/prometheus-route.yml
```diff
@@ -0,0 +1,13 @@
+apiVersion: route.openshift.io/v1
```
alexellis: We don't expose Prometheus in Kubernetes, should we for OpenShift?

openshift/openfaas-route.yml
```diff
@@ -0,0 +1,13 @@
+apiVersion: route.openshift.io/v1
```
alexellis: Would this make sense in the helm chart?

alexellis:
Thank you for the PR. I added 2x comments, please see the message from the bot.

openshift/prometheus-route.yml
mirroredge: I can remove that, in OpenShift there is not another way to access the Prometheus UI without a route the way it is deployed. Under most operations I would expect A user not needing to access it.

openshift/openfaas-route.yml
mirroredge: Ideally yes. OpenShift does not directly use helm, although it can be added on. Most people I have worked with in OpenShift don't use it either. I can research this more and add it in once I understand it, as I don't think this deployment option would hurt.

alexellis: OK, so perhaps we can make it optional if someone sets "openshift.createRoute: true" in their `values.yaml`?

alexellis: Also we generate YAML from our helm chart. So I think this should be there. I'll merge it now however, with a view to do the above.

alexellis:
cc @LucasRoesler

Issue #413, opened 2019-04-18
Switch to httpProbe for liveness/readiness


alexellis:
## Current Behaviour

The use of "exec" probes for liveness and readiness is required for software like the current incarnation of Istio, but can be very inefficient. [In Swarm this seems pronounced](https://github.com/moby/moby/issues/39102), but in Kubernetes less so.

## Possible Solution

HTTP Probes are now supported in the OpenFaaS watchdogs via `/_/health`. We should change over to httpProbes by default for liveness/readiness as soon as possible.

Some functions in the store may need to be rebuilt before the release is done and the helm chart is cut over to the new version to ensure that they have a compatible watchdog.

## Steps to Reproduce (for bugs)

1. This is available as a flag, added by VMware last year

## Context

Increasing efficiencies - especially at "idle"

There is also a chance that HTTP probes may reduce cold-start time from scaling from zero or rolling over to a new version of a function.

Once fully tested this needs to be repeated in the operator.

stefanprodan:
I see no harm in doing this if we have a flag in Helm that can let a Istio user to switch to exec. The http probes will work with Istio 1.2 see https://preliminary.istio.io/help/ops/setup/app-health-check/#probe-rewrite

LucasRoesler:
Seems like a good usecase for helm's template helpers to keep the deployment yaml clean

PR #414, opened 2019-04-18
Prepare for httpProbes

alexellis:

## Description

Prepare to switch to httpProbes and fix issues.

- httpProbe as an option is documented in the README

- In testing the update handler would not update the probe values
for deployments even if the controller was reloaded with new
config. This has been updated in the update handler.

- New method broken out for makeProbes to be used from
deploy/update.

## Motivation and Context

Fixes: #413

## How Has This Been Tested?

* Updated unit tests

* Scenario 1 - not updating the probes

Deployed faas-netes with current image and httpProbe=false, deployed cows. Then I updated the faas-netes deployment for httpProbe=true and redeployed cows, but the probes were not changed. With a new image from this PR this scenario was resolved.

* Scenario 2 - exec / http probes both work

In this example I tried toggling httpProbe on/off and was able to see it reflect in the function and the function was healthy. 

## Types of changes

## Checklist:

### DIFF
--- a/contrib/deploy.sh
+++ b/contrib/deploy.sh
@@ -5,8 +5,17 @@ export KUBECONFIG="$(kind get kubeconfig-path)"
 echo "Applying namespaces"
 kubectl apply -f ./namespaces.yml
 
-echo $(head -c 16 /dev/urandom| sha256sum | cut -d " " -f 1) > ./password.txt # Store password in password.txt file
-PASSWORD=$(cat ./password.txt)
+sha_cmd="sha256sum"
+if [ ! -x "$(command -v $sha_cmd)" ]; then
+    sha_cmd="shasum -a 256"
+fi
+
+if [ -x "$(command -v $sha_cmd)" ]; then
+    sha_cmd="shasum"
+fi
+
+PASSWORD=$(head -c 16 /dev/urandom| $sha_cmd | cut -d " " -f 1)
+echo -n $PASSWORD > password.txt
 
 kubectl -n openfaas create secret generic basic-auth \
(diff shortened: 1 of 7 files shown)


contrib/deploy.sh
```diff
@@ -5,8 +5,17 @@ export KUBECONFIG="$(kind get kubeconfig-path)"
 echo "Applying namespaces"
 kubectl apply -f ./namespaces.yml
 
-echo $(head -c 16 /dev/urandom| sha256sum | cut -d " " -f 1) > ./password.txt # Store password in password.txt file
-PASSWORD=$(cat ./password.txt)
+sha_cmd="sha256sum"
```
alexellis: @zeerorg FYI this broke on MacOS so I've put a fix in that @rgee0 added in the faas-cli.

alexellis:
@stefanprodan @LucasRoesler we'll need to apply this same or a similar patch to the openfaas-operator.

alexellis:
> This seems like a good opportunity to add a test for the probes configuration methods. Especially since we plan to make it mor complex soon.

Which method?

Make it more complex how?

LucasRoesler:
I am going to take a break from my computer tonight, because I can't seem to keep my sentences straight

alexellis:
Thanks for the suggestion. I've added a unit test for both cases (exec/httpProbe) and would be happy to add more later.

Have a good evening 👋

contrib/deploy.sh
zeerorg: I should've taken more care with sha256 utility, thanks for fixing this.

PR #469, opened 2019-07-04
Fix readFunction

alexellis:
## Description

Fix readFunction

## Details

Unfortunately some regression was caused by
a9554655badad60b01e41d6d9afcfeffca0e1b02 which caused OpenFaaS
Cloud not to show customer functions on their first commit
i.e. "create", but always after their second commit.

The reason was a difference between the logic for create/update
for managing labels.

Change was updated to only create labels for the Pod template,
but readFunction was left reading from the Deployment template
which meant that we had no labels.

The inconsistency was due to the update code adding the labels
where readFunction expected to find them.

This has been tested e2e and as a result:

readFunction - reads from the Pod template
create - writes to the Pod template
update - creates only in the Pod template

### DIFF
--- a/handlers/reader.go
+++ b/handlers/reader.go
@@ -89,7 +89,7 @@ func readFunction(item appsv1.Deployment) *requests.Function {
 		replicas = uint64(*item.Spec.Replicas)
 	}
 
-	labels := item.Labels
+	labels := item.Spec.Template.Labels
 	function := requests.Function{
 		Name:              item.Name,
 		Replicas:          replicas,
--- a/handlers/update.go
+++ b/handlers/update.go
@@ -3,11 +3,12 @@ package handlers
 import (
 	"encoding/json"
 	"fmt"
-	"github.com/openfaas/faas-netes/k8s"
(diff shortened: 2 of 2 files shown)


alexellis:
I'm paraphrasing @stefanprodan here. He said on Slack that his intent was to reduce the load for CoreDNS since each label is supposed to correspond to a DNS entry.

I don't think we strictly need labels at the Deployment level, but I prefer them there for filtering.

We do need consistency though and I would like to get this patched ASAP for the community cluster. Right now the first time a function is created, it is invisible 👎 

Alex

alexellis:
Thanks for the comment, it might be something we need to consider reverting.

I'm not sure how the migration scripts we have used for OpenFaaS Cloud would work that used label selectors on the Deployment for instance.

LucasRoesler:
Those would certainly have to be adjusted, but it would be possible because I think we would still have some labels on all deployments, e.g. `faas_function`, this would allow us to grab all Functions, at which point you can always filter on the annotations once you have them in memory.

Issue #484, opened 2019-08-02
Question: Why are functions updated when they are deployed?


bmcstdio:
## Expected Behaviour

Running `faas deploy` multiple times without changing anything doesn't cause a function (and hence its pods) to be re-created.

## Current Behaviour

Running `faas deploy` multiple times without changing anything causes a function (and hence its pods) to be re-created. This is (in part?) due to the value of the `UID` label being changed everytime, as well as to the fact that environment variables are not sorted:

https://github.com/openfaas/faas-netes/blob/master/handlers/update.go#L80
https://github.com/openfaas/faas-netes/blob/master/handlers/update.go#L71

## Possible Solution

I am not sure what the purpose of the `UID` label is, but maybe it could be removed. If that is not a possibility, then maybe we can provide an option that allows for overriding that value (and hence to keep it constant)? If this is acceptable as a solution, I believe I'll be able to work on a PR.

## Steps to Reproduce (for bugs)

1. Create a function and deploy it using `faas deploy`.
2. Run `faas deploy` again without changing anything.
3. Observe the pods being re-created.

## Context

My current thinking is that running `faas deploy` should be as idempotent as possible—meaning it is "safe" to run the command multiple times using the same parameters. This would be useful on CI/CD, for instance.

## Your Environment

* FaaS-CLI version ( Full output from: `faas-cli version` ):

```
(...)
CLI:
 commit:  750add9e2a6d1d1f1bfb788a6a85341b4b4140e8
 version: 0.8.22
```

* Docker version `docker version` (e.g. Docker 17.0.05 ):

```
Client: Docker Engine - Community
 Version:           19.03.1
 API version:       1.40
 Go version:        go1.12.5
 Git commit:        74b1e89
 Built:             Thu Jul 25 21:18:17 2019
 OS/Arch:           darwin/amd64
 Experimental:      false

Server: Docker Engine - Community
 Engine:
  Version:          19.03.1
  API version:      1.40 (minimum version 1.12)
  Go version:       go1.12.5
  Git commit:       74b1e89
  Built:            Thu Jul 25 21:17:52 2019
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          v1.2.6
  GitCommit:        894b81a4b802e4eb2a91d1ce216b8817763c29fb
 runc:
  Version:          1.0.0-rc8
  GitCommit:        425e105d5a03fabd737a126ad93d62a9eeede87f
 docker-init:
  Version:          0.18.0
[...]

PR #486, opened 2019-08-07
Sort env-vars by string name

alexellis:

## Description

Sort env-vars by string name

## Motivation and Context

This issue fixes the non-determinism observed by @LucasRoesler
and @bmcstdio which caused Function Pods to be re-created due
to the order of environment variables changing in the Deployment
and PodSpec.

Ref: https://github.com/openfaas/faas-netes/issues/484

## How Has This Been Tested?

Tested with unit tests

## Types of changes

### DIFF
--- a/handlers/secrets_test.go
+++ b/handlers/secrets_test.go
@@ -6,7 +6,7 @@ import (
 	"net/http"
 	"testing"
 
-	"github.com/openfaas/faas/gateway/requests"
+	types "github.com/openfaas/faas-provider/types"
 	appsv1 "k8s.io/api/apps/v1beta2"
 	apiv1 "k8s.io/api/core/v1"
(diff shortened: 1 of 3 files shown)


handlers/secrets_test.go
```diff
@@ -6,7 +6,7 @@ import (
 	"net/http"
 	"testing"
 
-	"github.com/openfaas/faas/gateway/requests"
+	types "github.com/openfaas/faas-provider/types"
```
bmcstdio: Nit: what's the reason for this change?

alexellis: A compilation error :-) this package moved, see the notes I posted in #faas-provider on Slack

bmcstdio:
@alexellis thank you for this, I can confirm it works (alongside with using a stable `uid` label)!

PR #509, opened 2019-09-19
Harden the security context of the OpenFaaS control-plane.

bmcustodio:
## Description

This PR hardens the security context of the OpenFaaS control-plane, while aligning the included pod security policy with the [`restricted` example pod security policy](https://kubernetes.io/docs/concepts/policy/pod-security-policy/#example-policies) (except for the AppArmor/SECCOMP annotations).

## Motivation and Context

- [ ] I have raised an issue to propose this change ([required](https://github.com/openfaas/faas/blob/master/CONTRIBUTING.md))

Running containers as a non-root user, disabling privilege escalation and generally restricting what containers can and cannot do is a standard best-practice.

## How Has This Been Tested?

I've rendered the chart using Helm, installed it and found no problems with the deployment.

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I've read the [CONTRIBUTION](https://github.com/openfaas/faas/blob/master/CONTRIBUTING.md) guide
- [x] I have signed-off my commits with `git commit -s`
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.

Issue #521, opened 2019-09-27
[Feature] Allow for customizing the security context of the OpenFaaS control-plane.


bmcustodio:
Following the discussion #509, I think it would be valuable to discuss how we could let users customize the security context of the OpenFaaS control-plane components without forcing any particular configuration on anyone. As far as I can tell, the only way of doing this would be to add a `.securityContexts` Helm value containing fields for every container and pod and possibly defaulting to the "safest" possible default value of `{}` (see #522).

alexellis:
We should have this now in the Profiles feature.

PR #522, opened 2019-09-27
Allow for customizing the security context of the OpenFaaS control-plane.

bmcustodio:
## Description

A follow-up to #521 and alternative to #509 (closes #509).

## Motivation and Context

The same as #509, but without modifying the included pod security policy and using, I believe, the "safest" default of all for each `securityContext` field, `{}`.

## How Has This Been Tested?

By rendering the templates, inspecting them for errors and installing the resulting manifests in a Minikube cluster.

## Types of changes

## Checklist:

### DIFF
--- a/chart/openfaas/README.md
+++ b/chart/openfaas/README.md
@@ -4,6 +4,30 @@
 
 [OpenFaaS](https://github.com/openfaas/faas) (Functions as a Service) is a framework for building serverless functions with Docker and Kubernetes which has first class support for metrics. Any process can be packaged as a function enabling you to consume a range of web events without repetitive boiler-plate coding.
 
+## Important notes
+
+If you are upgrading from pre-`6.0.0` and have `securityContext` set to `true`, please...
+
+* ... tweak the `securityContexts` field in `values.yaml` according to your requirements. By default, the new pods will have an empty `securityContext` field. To keep your existing configuration for this field, use
+
+  ```yaml
+  securityContexts:
+    basicAuthPlugin:
+      containers:
+        basicAuthPlugin:
+          readOnlyRootFilesystem: true
+          runAsUser: 10001
+    gateway:
+      containers:
+        faasNetes:
+          readOnlyRootFilesystem: true
+          runAsUser: 10001
+        gateway:
+          readOnlyRootFilesystem: true
+          runAsUser: 10001
+  ```
+* ... remove the `securityContext` field, which has been replaced by the aforementioned one.
+
 ## Highlights
 
 * Ease of use through UI portal and *one-click* install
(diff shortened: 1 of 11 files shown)


stefanprodan:
This looks great but it's a breaking change that will remove the security context if people will do `helm upgrade`. I think we should add migration documentation for those that are using `securityContext: true`. The chart version should get a major version bump to signal users that it contains breaking changes.

bmcstdio:
Thank you for your feedback @stefanprodan!

> I think we should add migration documentation for those that are using `securityContext: true`.

Agreed, I'll try to take care of that.

> The chart version should get a major version bump to signal users that it contains breaking changes.

I had made a minor version bump, but if you think a major bump is a better approach, I'll happily do that.

stefanprodan:
Helm charts should adhere to semantic versioning, a breaking change requires a major version bump see https://semver.org/

bmcstdio:
@stefanprodan bumped to `6.0.0` and added a note to `README.md`. Let me know how that looks.

chart/openfaas/README.md
```diff
@@ -4,6 +4,13 @@
 
 [OpenFaaS](https://github.com/openfaas/faas) (Functions as a Service) is a framework for building serverless functions with Docker and Kubernetes which has first class support for metrics. Any process can be packaged as a function enabling you to consume a range of web events without repetitive boiler-plate coding.
 
+## Important notes
+
+If you are upgrading from pre-`6.0.0`, please...
```
stefanprodan: This should tell people that have `securityContext: true` how to upgrade by keeping the same settings as they had before.

stefanprodan: The migration should include step like: 
- getting the current setting with `helm get values openfaas`
- creating a my-values.yaml with the the above values
- remove securityContext
- how to setup the new securityContexts to match the old settings
- how to upgrade using the custom values file

bmcstdio: Isn't "_tweak the `securityContexts` field (...) according to your requirements_" clear and flexible enough? Users that are updating may have different requirements now that they can in fact fully tweak the security context for their pods and containers, and I don't think we should be somehow suggesting that they should keep their existing settings for a successful update?

bmcstdio: @stefanprodan I don't think we should be telling users how to operate Helm, especially because said instructions can quickly get outdated, which makes things worse. I think that everyone that is using the Helm chart knows their way around Helm, and has certainly performed an update before. If not, they can certainly ask for help over Slack, don't you think?

stefanprodan: I think we should make them aware that without following the migration steps their OpenFaaS control plane will have **no security context** at all and show them how they can keep the current settings.

bmcstdio: @stefanprodan users that care about that most probably have their pod security policies in place, in which case they'll promptly notice their OpenFaaS control plane has "_*no security context* at all_". In any case, I've updated the `README.md` file in a way I believe is clear enough. Could you please check?

alexellis:
My understanding is that we added this feature originally to increase security,  but then refactored to use a toggle so that service mesh tooling could run, which needs root in the pod.  The read-only for system is a desirable default IMHO and doesn’t block service mesh. Do either of you have an objection or concern about that?

alexellis:
Rebase needed. Would like a reply to my question before merging.

bmcstdio:
> My understanding is that we added this feature originally to increase security, but then refactored to use a toggle so that service mesh tooling could run, which needs root in the pod. The read-only for system is a desirable default IMHO and doesn’t block service mesh. Do either of you have an objection or concern about that?

@alexellis sorry, but I do not follow. Could you please rephrase?

bmcstdio:
@alexellis could you provide some feedback, please?

alexellis:
The PR removes the readOnlyRootFilesystem from faas-netes and other components, but this is not desired. I would like to retain that default.

Screenshot 2019-10-17 at 12 09 04

Perhaps you could update the values file?

Screenshot 2019-10-17 at 12 08 57

Alex

bmcstdio:
@alexellis did you follow the discussion with @stefanprodan above? In any case, there is a warning about that in the README file.

alexellis:
Closing inactive PRs. If anyone needs this feel free to create a new issue.

PR #529, opened 2019-09-29
[WIP] Route on path when using proxy

alexellis:

## Description

Route on path when using proxy

## Motivation and Context

system-dashboard in OpenFaaS Cloud was getting an invalid path when Using faas-netes in pass-through mode for proxying to functions.

The change to use faas-provider means that functions can now
be invoked with a path. An additional change will be required
for any namespace parameter required.

## How Has This Been Tested?

With faas-federation and the OFC dashboard, it failed to work before due to invalid paths.

## Types of changes

### DIFF
--- a/handlers/lookup.go
+++ b/handlers/lookup.go
@@ -1,0 +1,30 @@
+// Copyright 2019 OpenFaaS Author(s)
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+package handlers
+
+import (
+	"fmt"
+	"net/http"
+	"net/url"
+)
+
+type FunctionLookup struct {
+}
+
+func (l *FunctionLookup) Resolve(name string) (url.URL, error) {
+	namespace := "openfaas-fn"
+
+	urlRes, errRes := url.Parse(fmt.Sprintf("http://%s.%s:8080", name, namespace))
+
+	return *urlRes, errRes
+}
+
+func NewFunctionLookup() http.HandlerFunc {
+	return func(w http.ResponseWriter, r *http.Request) {
+		defer r.Body.Close()
+
+		w.WriteHeader(http.StatusOK)
+
+	}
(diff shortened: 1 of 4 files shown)


handlers/lookup.go
```diff
@@ -0,0 +1,30 @@
...
+
+func (l *FunctionLookup) Resolve(name string) (url.URL, error) {
+	namespace := "openfaas-fn"
+
+	urlRes, errRes := url.Parse(fmt.Sprintf("http://%s.%s:8080", name, namespace))
+
+	return *urlRes, errRes
+}
+
+func NewFunctionLookup() http.HandlerFunc {
```
LucasRoesler: Why do we need this noop handlerfunc factory?

alexellis: You tell me.. I'm trying to implement the proxy you built and it seemed to be required. This fixes the issue, but if you can do your own PR it would be better.

alexellis: @LucasRoesler, I'm happy for you to take over this PR give that you built the proxy API in use here.

main.go
```diff
@@ -76,7 +78,7 @@ func main() {
 	factory := k8s.NewFunctionFactory(clientset, deployConfig)
 
 	bootstrapHandlers := bootTypes.FaaSHandlers{
-		FunctionProxy:        handlers.MakeProxy(functionNamespace, cfg.ReadTimeout),
+		FunctionProxy:        proxy.NewHandlerFunc(cfg.ReadTimeout, &handlers.FunctionLookup{}),
```
alexellis: @LucasRoesler do you mean that NewHandlerFunc only needs a `nil` parameter here?

LucasRoesler: No, I think this line is fine, i only mean that lines 23-30 in `handlers/lookup.go` aren't needed

Issue #511, opened 2019-09-20
[Feature] multiple namespace support


alexellis:
## Feature: multiple namespace support

The feature is outlined at https://github.com/openfaas/faas-netes/issues/433, which can be used for comments, questions, and suggestions.

The use-case is: multiple namespaces, single installation, i.e. `dev/staging`, `team-a/team-b`, `client-a/client-b`.

## Checklist:

* [x] Allow `.` in paths for faas-provider https://github.com/openfaas/faas-provider/pull/26
* [x] Allow `.` in paths for faas/gateway https://github.com/openfaas/faas/pull/1315
* [x] Update "direction functions" behaviour in gateway to allow namespace in path
* [x] Check/Update behaviour for auto-scaling via alerts https://github.com/openfaas/faas/pull/1316
* [x] Check the way Prometheus metrics are recorded
* [x] Define optional RBAC role for ClusterRole https://github.com/openfaas/faas-netes/commit/e2c013bfefbef9d224ce01ebb56f63b116d90ded
* [x] Update faas-netes CRUD to use namespace detected or passed in #513 #519
* [x] Update faas-netes to populate namespace in "get" endpoint - https://github.com/openfaas/faas-netes/pull/512
* [x] Update faas-netes, faas and faas-provider with list namespaces endpoint https://github.com/openfaas/faas/pull/1318 #518 https://github.com/openfaas/faas-provider/pull/27
* [x] Add a list functions flag / endpoint which enumerates all valid namespaces
* [x] Add namespace filtering to only OpenFaaS namespaces
* [x] Update CLI to become namespace-aware for Function CRUD https://github.com/openfaas/faas-cli/pull/694
* [x] Check/Update [nats-queue-worker](https://github.com/openfaas/nats-queue-worker) so that it is namespace-aware 
* [x] Check/Update [connector-sdk](https://github.com/openfaas-incubator/connector-sdk/) so that it is namespace-aware https://github.com/openfaas-incubator/connector-sdk/pull/36
* [x] Document approach for testing e2e
* [x] Update docs repo where required to document multiple-namespaces 

CLI

* [x] Deploy via `--namespace` for single function
* [x] Deploy via stack.yml with multiple functions - https://github.com/openfaas/faas-cli/issues/737
UI https://github.com/openfaas/faas/issues/1379

* [x] UI - add namespace selector
* [x] UI - add "all namespaces" option

Secrets

* [x] Update faas-netes for CRUD for secrets to use namespace detected or passed 

[...]

Issue #28, opened 2017-09-22
Enable variable timeouts for HTTP


alexellis:
This matches the API gateway in the faas repo:

```
read_timeout - seconds as an integer
write_timeout - seconds as an integer
```

The code can be copied from read_config.go

e-beach:
Doing.

Note that the code Alex is referring to is at https://github.com/alexellis/faas/blob/1aeb978617258da80a684121ce8b3e72e7201910/gateway/types/readconfig.go#L56-L57

alexellis:
@e-beach are you working on this?

alexellis:
Derek add label: hacktoberfest

PR #530, opened 2019-10-03
Add namespace flag to secrets handlers

jonatasbaldin:
## Description
Add support to Namespace in Secrets.

I created the `getLookupNamespace` function, which I'm not 100% proud of, since it reads the `r.Body` and puts it back into the `r` struct. BUT this way I didn't need to change all the signatures of all the other handlers, also, if there's any verification for the namespace we can do in one place.

Depends on (build will fail before merging the PRs below and bumping their version in this project):
- [faas-provider #28](https://github.com/openfaas/faas-provider/pull/28).
- [faas #1334](https://github.com/openfaas/faas/pull/1334).

## Motivation and Context

- [x] I have raised an issue to propose this change ([required](https://github.com/openfaas/faas/blob/master/CONTRIBUTING.md)) [faas-netes #511](https://github.com/openfaas/faas-netes/issues/511).

## How Has This Been Tested?
Added tests ✨ Also I deployed locally and tested with the following commands:
```bash
$ faas-cli secret create secret-cli --from-literal=value

$ curl -u admin:$PASSWORD $OPENFAAS_URL/system/secrets
[{"name":"secret-cli","namespace":"openfaas-fn"}]

# no secret in the default ns
$ curl -u admin:$PASSWORD $OPENFAAS_URL/system/secrets\?namespace\=default    
[]

# creates a secret in the default ns
$ curl -u admin:$PASSWORD $OPENFAAS_URL/system/secrets -X POST -d '{"name": "secret-api", "value": "value", "namespace": "default"}' -H 'Content-Type: application/json

# doesn't list the default secret 
$ curl -u admin:$PASSWORD $OPENFAAS_URL/system/secrets                    
[{"name":"secret-cli","namespace":"openfaas-fn"}]

# list the default secret
$ curl -u admin:$PASSWORD $OPENFAAS_URL/system/secrets\?namespace\=default
[{"name":"secret-api","namespace":"default"}

# changes the secret value from default ns
$ curl -u admin:$PASSWORD $OPENFAAS_URL/system/secrets -X PUT -d '{"name": "secret-api", "value": "value-PUT", "namespace": "default"}' -H 'Content-Type: application/j
son'

# verify secret value
$ k get secret secret-api -o jsonpath='{.data.secret-api}' | base64 -D
value-PUT

# delete secret
$ curl -u admin:$PASSWORD $OPENFAAS_URL/system/secrets -X DELETE -d '{"name": "secret-api", "namespace": "defa
ult"}' -H 'Content-Type: application/json' 

# verify deletion of `secret-api`
$ faas-cli secret list

NAME
secret-cli
```

Yay! ❤️

## Types of changes

## Checklist:

### DIFF
--- a/handlers/secrets.go
+++ b/handlers/secrets.go
@@ -4,6 +4,7 @@
 package handlers
 
 import (
+	"bytes"
 	"encoding/json"
 	"fmt"
 	"io"
@@ -28,28 +29,75 @@ const (
 
 // MakeSecretHandler makes a handler for Create/List/Delete/Update of
 //secrets in the Kubernetes API
-func MakeSecretHandler(namespace string, kube kubernetes.Interface) http.HandlerFunc {
+func MakeSecretHandler(defaultNamespace string, kube kubernetes.Interface) http.HandlerFunc {
 	return func(w http.ResponseWriter, r *http.Request) {
 		if r.Body != nil {
 			defer r.Body.Close()
 		}
 
-		client := kube.CoreV1().Secrets(namespace)
+		lookupNamespace, err := getLookupNamespace(defaultNamespace, kube, r)
+		if err != nil {
+			switch err.Error() {
+			case "unable to unmarshal Secret request":
+				http.Error(w, err.Error(), http.StatusBadRequest)
+			case "unable to manage secrets within the specified namespace":
+				http.Error(w, err.Error(), http.StatusUnauthorized)
+			default:
+				w.WriteHeader(http.StatusBadRequest)
+				return
+			}
+		}
+
+		client := kube.CoreV1().Secrets(lookupNamespace)
 
 		switch r.Method {
 		case http.MethodGet:
 			getSecretsHandler(client, w, r)
 		case http.MethodPost:
-			createSecretHandler(client, namespace, w, r)
+			createSecretHandler(client, lookupNamespace, w, r)
 		case http.MethodPut:
-			replaceSecretHandler(client, namespace, w, r)
+			replaceSecretHandler(client, lookupNamespace, w, r)
 		case http.MethodDelete:
-			deleteSecretHandler(client, namespace, w, r)
+			deleteSecretHandler(client, lookupNamespace, w, r)
 		default:
 			w.WriteHeader(http.StatusBadRequest)
 			return
 		}
 	}
+}
+
+// getLookupNamespace returns the namespace specified in any request type
+func getLookupNamespace(defaultNamespace string, kube kubernetes.Interface, r *http.Request) (string, error) {
+	req := types.Secret{Namespace: defaultNamespace}
+
+	switch r.Method {
+	case http.MethodGet:
+		q := r.URL.Query()
+		namespace := q.Get("namespace")
+
+		if len(namespace) > 0 {
+			req.Namespace = namespace
+		}
+
+	case http.MethodPost, http.MethodPut, http.MethodDelete:
+		body, _ := ioutil.ReadAll(r.Body)
+		err := json.Unmarshal(body, &req)
+		if err != nil {
+			log.Printf("error while getting namespace: %s\n", err)
+			return "", fmt.Errorf("unable to unmarshal Secret request")
+		}
+
+		// Reconstruct Body
+		r.Body = ioutil.NopCloser(bytes.NewBuffer(body))
+	}
+
+	allowedNamespaces := list(defaultNamespace, kube)
+	ok := findNamespace(req.Namespace, allowedNamespaces)
+	if !ok {
+		return req.Namespace, fmt.Errorf("unable to manage secrets within the %s namespace", req.Namespace)
+	}
+
+	return req.Namespace, nil
 }
 
 func getSecretsHandler(kube typedV1.SecretInterface, w http.ResponseWriter, r *http.Request) {
@@ -65,7 +113,8 @@ func getSecretsHandler(kube typedV1.SecretInterface, w http.ResponseWriter, r *h
 	secrets := []types.Secret{}
 	for _, item := range res.Items {
 		secret := types.Secret{
-			Name: item.Name,
+			Name:      item.Name,
+			Namespace: item.Namespace,
 		}
 		secrets = append(secrets, secret)
 	}
--- a/handlers/secrets_api_test.go
+++ b/handlers/secrets_api_test.go
@@ -13,6 +13,7 @@ import (
 	"testing"
 
 	"github.com/openfaas/faas/gateway/requests"
+	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	testclient "k8s.io/client-go/kubernetes/fake"
 )
@@ -21,8 +22,6 @@ func Test_SecretsHandler(t *testing.T) {
 	namespace := "of-fnc"
 	kube := testclient.NewSimpleClientset()
 	secretsHandler := MakeSecretHandler(namespace, kube).ServeHTTP
-	secretLabel := "app.kubernetes.io/managed-by"
-	secretLabelValue := "openfaas"
 	secretName := "testsecret"
 
 	t.Run("create managed secrets", func(t *testing.T) {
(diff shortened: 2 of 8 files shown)


handlers/secrets.go
```diff
@@ -28,30 +29,70 @@ const (
...
+		// Reconstruct Body
+		r.Body = ioutil.NopCloser(bytes.NewBuffer(body))
+
+		if req.Namespace != "" {
+			lookupNamespace = req.Namespace
+		}
+	}
+
+	if lookupNamespace == "kube-system" {
+		return "", fmt.Errorf("unable to list within the kube-system namespace"), http.StatusUnauthorized
```
viveksyngh: I think we can update this error message to something like .. `unable to manage secrets within the kube-system namespace`. which is related to what user is trying to do.

jonatasbaldin: Agreed and changed 💃

alexellis:
> I created X which I'm not 100% proud of since Y

^ Hmm, is there an alternative way you could do it then? It sounds like you are not convinced that it's the right solution? cc @LucasRoesler

handlers/secrets.go
alexellis: Yes, but for every place it occurs, not just the one message.

alexellis:
```
 curl -u admin:$PASSWORD $OPENFAAS_URL/system/secrets\?namespace\=default
[{"name":"secret-api","namespace":"default"}
```

Given that the namespace is an input, and we can only have one, why would you say we need it as an output? I'm not saying it's right or wrong, I'm curious about the design.

jonatasbaldin:
> > I created X which I'm not 100% proud of since Y
> 
> ^ Hmm, is there an alternative way you could do it then? It sounds like you are not convinced that it's the right solution? cc @LucasRoesler

I'm ok with this solution. Another one would be to do the query parameter verification in the GET handler and the POST verification in the other parameter, which would involve generate the kube client inside each parameter, which would change a lot.

But I think the solution is ok enough to not justify all these changes.

Would love ur input @LucasRoesler!

jonatasbaldin:
> [{"name":"secret-api","namespace":"default"}

jonatasbaldin:
> ```
>  curl -u admin:$PASSWORD $OPENFAAS_URL/system/secrets\?namespace\=default
> [{"name":"secret-api","namespace":"default"}
> ```
> 
> Given that the namespace is an input, and we can only have one, why would you say we need it as an output? I'm not saying it's right or wrong, I'm curious about the design.

I added because in a client – like the CLI – is probably easier to build return data listing the secret name and namespace if the result is in the body.

handlers/secrets.go
```diff
@@ -28,30 +29,70 @@ const (
...
+	if r.Method == http.MethodGet {
+		q := r.URL.Query()
+		namespace := q.Get("namespace")
+
+		if len(namespace) > 0 {
+			lookupNamespace = namespace
+		}
+	} else {
+		body, _ := ioutil.ReadAll(r.Body)
+		req := types.Secret{}
```
LucasRoesler: you could also use an anonymous struct here, which will possibly make it more flexible

```go
req := struct{ Namespace string }{}
```

This method could then be used for any POST/PUT/PATCH request that has a `Namespace` field in the body.

In that case, you could use a switch statement on `r.Method`  to handle all of these cases

```go
switch r.Method {
case http.MethodPost, http.MethodPut, http.Patch:
    // load req body
default:
    // load from GET param
}
```

handlers/secrets.go
```diff
@@ -28,30 +29,70 @@ const (
...
+			deleteSecretHandler(client, lookupNamespace, w, r)
 		default:
 			w.WriteHeader(http.StatusBadRequest)
 			return
 		}
 	}
 }
 
+// getLookupNamespace returns the namespace specified in any request type
+func getLookupNamespace(defaultNamespace string, r *http.Request) (string, error, int) {
```
LucasRoesler: Another modification that would generalize this is to convert it to an interface and put the `defaultNamespace` and excluded namespaces into a struct.  This would look like 

```go
type NamespaceResolver interface {
    Lookup(r *http.Request) (string, error, int)
}

type namespaceResolver struct {
    defaultNamespace string
    exclude map[string]bool
}

func NewNamespaceResolver(defaultNamespace string, exclude ...string) NamespaceResolver {
    r = namespaceResolver{
        defaultNamespace: defaultNamespace
        exclude: map[string]bool{"kube-system": true}
    }

    for _, n := range exclude {
        r.exclude[n] = true
    }

    return &r
}

func (r *namespaceResolver) Lookup(r *http.Request) (string, error, int) {
    //
    return "", nil, 0
}
```

This allows use to extend the excluded list, for example via a flag during startup.  Also, using a `map[string]bool` allows you to keep the exclusion check very simple

```go
if _, ok := r.exclude[lookupNamespace]; ok {
    return "", fmt.Errorf("unable to manage secrets within the %s namespace", lookupNamespace), http.StatusUnauthorized
}
```

For now the the namespace resolver would simply be initialized  without any extra exclude values:

```go
namespaceResolver := NewNamespaceResolver(defaultNamespace)
```

Which could happen either at the start of the `MakeSecretHandler` or could be initialized in the server setup and then passed in.


_If_ we opted to pass the resolver to the Handler, then it would look roughly like this

```go
func MakeSecretHandler(namespaceResolver NamespaceResolver, kube kubernetes.Interface) http.HandlerFunc { }
```
Passing it in will ensure that it is easy to mock in tests and the behavior can be defined and controlled globally for the server.

There is a lot to digest here, i am not saying we must do it this way.  Just a suggestion to consider.

handlers/secrets.go
jonatasbaldin: I like this, will implement!

jonatasbaldin: I like this, will implement!

handlers/secrets.go
jonatasbaldin: Although I like the solution, I think it becomes too complex for now, since we are just ignoring one specific namespace. IMHO using a simple `getLookupNamespace` makes it easier to reason about the code.

Shall we go the way I coded it?

LucasRoesler: That is fine, but I was thinking more on this as well and there are two other aspects that we should address before we merge.  I will add those as separate comments.

handlers/secrets.go
```diff
@@ -28,30 +29,67 @@ const (
...
```
LucasRoesler: In general, I would prefer the handler to decide which error code to return instead.  One, I generally expect `error` to be the last item returned from a method

```go
func getLookupNamespace(defaultNamespace string, r *http.Request) (string, error) { }
```

And second, I think the separation of concerns is cleaner when a handler catches errors and then decides what the appropriate code is. 

This method could return a sentinel error (see https://dave.cheney.net/2019/06/10/constant-time for an example if you are not familiar) for each error case and then have the handler can easily use a switch to decide the correct status code.  This would then be cleaner for us to reuse the code in the future in cases that are not handlers, for example, in the openfaas-operator

alexellis:
@LucasRoesler thank you for the abstraction suggestion. I would like to get logs and secrets working e2e and then for someone to refactor with an abstraction like this afterwards in a follow-up PR.

alexellis:
I prefer to merge PRs that do only one thing, i.e. one from below:

* Add namespace flag to secrets commands
* Refactor all commands to abstraction pattern

handlers/secrets.go
```diff
@@ -28,30 +29,67 @@ const (
...
+		if err != nil {
+			log.Printf("error while getting namespace: %s\n", err)
+			return "", fmt.Errorf("unable to unmarshal Secret request"), http.StatusBadRequest
+		}
+
+		// Reconstruct Body
+		r.Body = ioutil.NopCloser(bytes.NewBuffer(body))
+	}
+
+	if req.Namespace == "kube-system" {
```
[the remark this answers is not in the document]
jonatasbaldin: I think this is yet another issue. We have checks for the `kube-system` namespace in various place, would need a slightly bigger refactor.

handlers/secrets.go
jonatasbaldin: Thanks for the advice and the article, really good read :)

I refactored the code so the handler catches the error, based on the error string (_is there a better way than checking the string_?).

handlers/secrets.go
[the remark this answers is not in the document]
LucasRoesler: I am not sure I agree. This change allows the user to modify secrets in namespaces that it isn't allowed to deploy functions to. There is already code for getting and checking the list of allowed. The refactor here would be to call the `list` method and then, in the `if` block, use the `findNamepsace` method. In fact, you could combine those into a single statement, so it could even be a one line change to the PR.

Refactoring the other checks can definitely be done in a separate PR, but I don't think we should ignore it here especially when there is a drop in solution already implemented.

But, I am ok with merging this if @alexellis wants to get the functionality in first and refactor later

handlers/secrets.go
```diff
@@ -65,7 +111,8 @@ func getSecretsHandler(kube typedV1.SecretInterface, w http.ResponseWriter, r *h
 	secrets := []types.Secret{}
 	for _, item := range res.Items {
 		secret := types.Secret{
-			Name: item.Name,
+			Name:      item.Name,
+			Namespace: item.Namespace,
```
LucasRoesler: This line is incorrect and is why the tests are currently failing, the `faas-provider` needs to be updated, the latest version is `0.12.0`. This needs to be updated in the `Gopkg.toml` and then you need to run `dep ensure` again 


```go
// Secret for underlying orchestrator
type Secret struct {
	Name  string `json:"name"`
	Value string `json:"value,omitempty"`
}
```

handlers/secrets.go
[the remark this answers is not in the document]
jonatasbaldin: Oh yes, I meant opening a PR for the other areas, I'll quickly add the check for this function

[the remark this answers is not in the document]
LucasRoesler: Awesome and thanks!

handlers/secrets.go
alexellis: @jonatasbaldin PTAL

handlers/secrets.go
```diff
@@ -28,30 +29,75 @@ const (
...
 		default:
 			w.WriteHeader(http.StatusBadRequest)
 			return
 		}
 	}
 }
 
+// getLookupNamespace returns the namespace specified in any request type
+func getLookupNamespace(defaultNamespace string, r *http.Request) (string, error) {
+	req := struct{ Namespace string }{Namespace: defaultNamespace}
```
alexellis: I am finding this confusing. I know Lucas suggested it, but why do we need this over a named type?

handlers/secrets_api_test.go
```diff
@@ -198,3 +197,71 @@ func Test_SecretsHandler_ListEmpty(t *testing.T) {
 		t.Errorf(`want empty list to be valid json i.e. "[]", but was %q`, string(body))
 	}
 }
+
+func Test_GetLookupNamespace(t *testing.T) {
+	defaultNamespace := "openfaas-fn"
```
alexellis: Would this work as a test-case table instead? Example: https://blog.alexellis.io/golang-writing-unit-tests/

handlers/secrets.go
LucasRoesler: This allows you to unmarshal any request that contains a Namespace field, and not  just one particular request type.

jonatasbaldin:
This last commit has the changes needed to check the allowed namespace list.

First, I needed to change the signature from the `MakeNamespacesLister` and `list` methods from `namespaces.go`. They were expecting a concrete type, and I couldn't use the `kube` client from `secrets.go`. Now they are the `kubernetes.Interface` ✨

And in the tests, I'm testing against an allowed namespace – created with the `openfaas` annotation – and a forbidden one, which has no annotation.

handlers/secrets.go
jonatasbaldin: Done.

handlers/secrets.go
jonatasbaldin: I think mainly because we didn't have the `Secret` type from `faas-provider` with the new `Namespace` field. Now we do. Will update it.

handlers/secrets_api_test.go
jonatasbaldin: I can look into refactoring into table tests, let's just get everything right first 💃

alexellis:
We can't take a merge commit via the rebase command, please can you revert that last change and apply it with a rebase instead?

```
git pull origin master --rebase
```

alexellis:
In other words every commit has to be signed off and merge commits cannot be signed off.

Undo the merge:
```
git reset HEAD~1
```

Rebase from master

```
git pull origin master --rebase
```

Fix the conflicts:

```
# Edit the files
git add .
git rebase --continue
```

Then push back up

```
git push origin --force
```

jonatasbaldin:
Thanks @alexellis, I hope this output is the expected (looks like a mess) hehehe

alexellis:
Why does it look like a mess? No, I was just trying to help you revert your unsigned merge commit.

alexellis:
I think we are ready to squash these commits into one and then merge, that's something you can do at your end @jonatasbaldin 

Btw I can't see why this "is a mess"? @LucasRoesler thoughts?

jonatasbaldin:
No worries man. I was joking because the PR looked a mess after the rebase because all the commits were listed again, I didn't know this was a thing. Just a :joy: reaction to the output hehehe

Everything is fine 🤙

I can squash the commits a bit later, ok?

Very happy we are almost merging this 🎉✨

jonatasbaldin:
Squashed ✅

LucasRoesler:
If we merge this first, I will rebase my PR https://github.com/openfaas/faas-netes/pull/532 afterwards because there are a couple of places that they overlap

alexellis:
Thank you for your patience @jonatasbaldin and for the awesome PR to faas-netes. Keep them coming

And when you get a moment, checkout the ask on the unit test table, does it make sense?

jonatasbaldin:
Thanks for all the feedback @alexellis and @LucasRoesler ❤️

@alexellis will put in my backlog for this week :)

alexellis:
Thank you @jonatasbaldin

PR #538, opened 2019-10-24
Ensure that static manifest inherit the default values

LucasRoesler:
## Description

- Use both the default values and the arm sepecific values files so that
  defaults can be cleanly defined in the main chart and inherited when
  creating the static manifests

## Motivation and Context

This came from a discussion with Alex while trying to debug something in the chart and realizing that the arm manifests might not be inheriting all of the default values.  This should allow him to test and verify 

## Types of changes

## Checklist:

### DIFF
--- a/contrib/create-static-manifest.sh
+++ b/contrib/create-static-manifest.sh
@@ -16,6 +16,7 @@
 
 CHART_DIR=${1:-"./chart/openfaas"}
 OUTPUT_DIR=${2:-"./yaml"}
+BASEVALUESNAME="$CHART_DIR/values.yaml"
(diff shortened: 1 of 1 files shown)


alexellis:
Is this PR to bring defaults into the armhf values file as a kind of merge?

LucasRoesler:
This ensures that any default values set in the default values will always propagate when we do the arm generation.

alexellis:
OK so I would want this to be tested on an ARM device before merging to check there is no unintended knock-on effect from taking in those values. 

I can do that unless someone else in the community wants to try it.

alexellis:
I don't have time to test this. Please ask in #arm-and-pi on Slack for help.

PR #561, opened 2019-12-04
Add nodeSelector for CPU architecture

alexellis:

## Description

Add nodeSelector for CPU architecture

## Motivation and Context

This change means that OpenFaaS can be installed into a mixed
architecture Kubernetes cluster.

Different architectures are added to the various values*.yaml files for armhf/arm64/amd64, @carlosedp may need a separate override for Risc-V and those using PowerPC should also override this, perhaps with an empty value?

## How Has This Been Tested?

Tested with k3d by applying the local chart, the nodeSelector was printed as expected and the containers came up / were scheduled.

## Types of changes

This may break for some people using and esoteric CPU architecture.


carlosedp:
Thanks for the mention Alex, I'm keeping a separate set of yaml files (on Kubernetes) and the docker-compose file as well for Risc-V.
I haven't submitted a PR yet since there are no official images for OpenFaaS for it yet.

If people is interested it's in https://github.com/carlosedp/riscv-bringup/tree/master/OpenFaaS for Docker Swarm and https://github.com/carlosedp/riscv-bringup/tree/master/kubernetes/openfaas for Kubernetes.

alexellis:
Is it not currently enough to maintain a values-riscv.yaml, or do we need to improve the chart a little?

carlosedp:
We could have a values-riscv64.yaml for sure but the images need to come from my Dockerhub since OpenFaaS CI still doesn't generate them. Also Go isn't still upstreamed with Risc-V support so default Go wouldn't be able to generate the binaries.

PR #579, opened 2020-01-07
chart: let helm know that CRDs are installed.

Gottox:
This change lets both, helm3 and helm2 know that CRDs are installed and adds special handling for those cases

## Description
Helm has special handling for CRDs that ensures that the API Server knows about those resources before helm continues the installation. 

## Motivation and Context
This makes it possible to deploy helm charts that contain the resource definitions as well as resources of this kind.

## How Has This Been Tested?
This change was tested using `helm template` (both helm2 and helm3) and diff the output to the output of master

## Types of changes

## Checklist:


alexellis:
@munnerz thoughts on this? I believe that cert-manager stays well away from this sort of thing?

stefanprodan:
The CRDs support in Helm v3 is seriously flawed, once you move the CRDs to the `crds` dir, Helm v3 will ignore them during upgrades making it impossible to ship updates for CRDs, see https://github.com/helm/helm/issues/6581

SpaWn2KiLl:
> The CRDs support in Helm v3 is seriously flawed, once you move the CRDs to the `crds` dir, Helm v3 will ignore them during upgrades making it impossible to ship updates for CRDs, see [helm/helm#6581](https://github.com/helm/helm/issues/6581)

On Helm [documentation](https://helm.sh/docs/chart_best_practices/custom_resource_definitions/):
> There is no support at this time for upgrading or deleting CRDs using Helm. This was an explicit decision after much community discussion due to the danger for unintentional data loss. Furthermore, there is currently no community consensus around how to handle CRDs and their lifecycle. As this evolves, Helm will add support for those use cases.

This behavior seems to be intentional so I don't see this has an issue for this PR acceptance.

alexellis:
@stefanprodan what would your recommendation be now?

Gottox:
I strongly disagree with @SpaWn2KiLl here. This is the way CRDs are supposed to work with helm3.

Any operator that is familiar enough with helm can fix the resulting issue. On the other hand, if openfaas decides to differ from the standard this will result in unexpected behavior from openfaas and in the worst case result in dataloss, which is the reason for this behavior in the first place.

SpaWn2KiLl:
There seems to be two issues right now on Openfaas regarding to CRD installations:
- `Profile` CRD is being installed on every openfaas instance. This is a problem when you need more than one release of the chart installed on the same cluster. Issue #686;
- An user that doesn't have Cluster Admin role with permissions for creating new CRDs can't install a release of the Openfaas chart since it tries to create the CRD on every installation.

The first issue could be addressed in two different ways:
- Creating a new flag value that specifies whether the CRD should be installed or not;
- Moving the CRD definition to `crd` folder which is the helm native way to install CRDs.

The second solution at first sounds to be the best strategy but then there comes the problem where you can't upgrade the CRD definition via `helm upgrade`. I'm not sure though how often will Openfaas CRDs need to be upgraded for a new versions.

But the [second issue](https://github.com/openfaas/faas-netes/issues/686#issuecomment-718586934) which is not having enough permissions can't be solved by simply moving CRDs to `crd` folder. I don't even know if helm can check if the CRD in `crd` folder is already installed if the operator that is trying to install the chart doesn't have permissions to list cluster's installed CRDs.

Anyway, I think the best solution would be simply flag all CRDs because it will be possible to:
- Install CRDs if they aren't installed yet
- Update CRDs if they are already installed
- Skip CRDs installation if the user doesn't have cluster permissions for that

The biggest issue right now is #686. That was a breaking change on newest Openfaas major. This prevents to have multiple releases installed on the namespace.

munnerz:
My advice remains the same - if you want to be able to update your CRD *at all* in a future release, you should avoid using Helm's `crd/` directory as you will be giving up the ability to evolve your CRD resources in any way.

My concrete advice would be to either:

1) keep the CRDs as they are in the `templates/` directory, although yes this *will* mean it'll be installed for each instance of the chart, potentially causing conflicts if you expect users to be able to install your chart multiple times in a single cluster
2) create a separate Helm chart that contains just your CRDs (and any supporting components like conversion webhooks etc) and instruct users to install *that* chart first (which will then allow them to install the 'main' chart as many times as is required).

> I'm not sure though how often will Openfaas CRDs need to be upgraded for a new versions.

From a maintainer PoV, I'd be *really* hesitant to just assume that you'll _never_ want to change your CRD resource.

SpaWn2KiLl:
I would go with:
> keep the CRDs as they are in the templates/ directory, although yes this will mean it'll be installed for each instance of the chart, potentially causing conflicts if you expect users to be able to install your chart multiple times in a single cluster

But the CRD would need to have a flag to specify whether the CRD should be installed or not, like `Function` CRD has.

ArcticXWolf:
So the easiest solution would be to just surround the Profiles CRD with a flag from the values, wouldn't it?

alexellis:
@ArcticXWolf yes, please have a look at #696 I think that unblocks us all and we can close this draft PR?

PR #577, opened 2020-01-06
Generating openfaas.yaml with helm 3

mkoeper:
template syntax changed slightly for helm 3

## Description
New syntax as per helm template --help:
Usage:
  helm template [NAME] [CHART] [flags]

## Motivation and Context
Ran into an error when following the instructions:
```
$ helm template faas-netes/chart/openfaas \
>     --name openfaas \
>     --namespace openfaas  \
>     --set basic_auth=true \
>     --set functionNamespace=openfaas-fn > $HOME/openfaas.yaml
Error: unknown flag: --name
```
- [X] I have raised an issue to propose this change ([required](https://github.com/openfaas/faas/blob/master/CONTRIBUTING.md))

## How Has This Been Tested?
Successful command was:
```
$ helm template openfaas faas-netes/chart/openfaas \
     --namespace openfaas \
     --set basic_auth=true \
     --set functionNamespace=openfaas-fn > $HOME/openfaas.yaml
```

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [x] Updated documentation

## Checklist:

- [ ] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] I've read the [CONTRIBUTION](https://github.com/openfaas/faas/blob/master/CONTRIBUTING.md) guide
- [ ] I have signed-off my commits with `git commit -s`
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.

Issue #520, opened 2019-09-24
[Feature] Add support for Helm3


alexellis:
## Feature

Add support for Helm3.

Helm3 removes the need for Tiller, which is considered insecure.

https://v3.helm.sh

## Context

On the members team call Stefan Prodan @stefanprodan volunteered to add Helm3 support to the OpenFaaS chart, some additional changes are required around arrangement of files such as the CRDs.

@LucasRoesler volunteered to test / verify the changes.

jonatasbaldin:
hey there!

@stefanprodan if you need any help we can pair on this 🎉  I'm also available for testing!

stefanprodan:
@jonatasbaldin can you please test that the current chart installed with `operator.createCRD=false` works with Helm v3? 

PS. See https://helm.workshop.flagger.dev/prerequisites/#helm-v3 on how to install Helm v3 CLI without overriding your v2 setup.

jonatasbaldin:
hey @stefanprodan 

I'm testing using K8S in Docker for Desktop.

## Test 1 (failed):
Installing default OpenFaaS with `operator.create=true` and `operator.createCRD=false`:
```
$ helm3 install openfaas openfaas/openfaas --namespace openfaas --set basic_auth=true --set functionNamespace=openfaas-fn --set operator.create=true --set operator.createCRD=false

# login...

$ faas list
Server returned unexpected status code: 500 - Error pulling metrics from provider/backend. Status code: 400
```

## Test 2 (failed):
Installing default OpenFaaS and upgrading to `operator.create=true` and `operator.createCRD=false`:
```
$ helm3 install openfaas openfaas/openfaas --namespace openfaas --set basic_auth=true --set functionNamespace=openfaas-fn

# login...

$ faas list
Function                      	Invocations    	Replicas

# upgrading to `operator.create=true` and `operator.createCRD=false`
$ helm3 upgrade openfaas openfaas/openfaas -n openfaas --set operator.create=true  --set operator.createCRD=false

$ faas version
  ___                   _____           ____
 / _ \ _ __   ___ _ __ |  ___|_ _  __ _/ ___|
| | | | '_ \ / _ \ '_ \| |_ / _` |/ _` \___ \
| |_| | |_) |  __/ | | |  _| (_| | (_| |___) |
 \___/| .__/ \___|_| |_|_|  \__,_|\__,_|____/
      |_|

CLI:
 commit:  893972afffa5ea6672faa6e11743ae46322ec58c
[...]

PR #583, opened 2020-01-14
Add helm3 usage to chart README

alexellis:

## Description

Add helm3 usage to chart README

## Motivation and Context

The easiest way to install is via:
"k3sup app install openfaas -helm3", however this brings parity
to the helm README file and HELM.md install instructions with
k3sup.

Tested against a k3d cluster, including the deletion command
which is different. Beware that helm uses namespaces to store
its own configuration.

Closes #577 and ref: #520

### DIFF
--- a/HELM.md
+++ b/HELM.md
@@ -13,9 +13,7 @@ curl -sSLf https://raw.githubusercontent.com/helm/helm/master/scripts/get-helm-3
 
 That's it, no tiller is required for using helm3.
 
-Now deploy OpenFaaS via Helm:
-
-How to install chart please follow install instructions in chart's [readme](chart/openfaas/README.md).
+Now deploy OpenFaaS via Helm: using the [Chart's readme](chart/openfaas/README.md).
 
 ## Helm 2 (legacy)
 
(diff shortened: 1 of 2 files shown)


HELM.md
```diff
@@ -1,7 +1,23 @@
 # Helm
 
 [Helm](https://github.com/kubernetes/helm) is a client CLI used to deploy Kubernetes
-applications. It supports templating of configuration files, but also needs a server
+applications.
+
+## Helm 3
```
LucasRoesler: I wonder if it would be cleaner to have a `HELM.md` and `HELM2.md` instead of branching so many sections of the document. 

The current version of helm _is_ helm 3, we should document that and provide a link to existing instructions for helm 2.  Keeping "legacy" instructions inline like this seems unnecessary.

The default install via brew is helm 3 and the first result in google is the install instructions for helm 3.  The quickstart linke on line 23 (below this) even points to the helm 3 docs.  We could include a "not sure of your version, run `helm version`  to check" statement.

alexellis: It already feels like a lot of indirection having the docs go to HELM.md then to charts/openfaas/README.md, having HELM2.md and HELM.md might be more modular, but it could also be a lot of clicks.

I don't know what the perfect version of this is, but let's start with what's there. Thank you for these suggestions.

alexellis: Bear in mind that most of our users (that I'm aware of) are still using helm 2, as are other parts of our docs, blog posts, tutorials and maybe the workshop. I suspect helm3 will be a more graduated move than a strict lift and shift. (I might be wrong)

PR #584, opened 2020-01-21
Convert faas-netes to Go modules

alexellis:

## Description

Convert faas-netes to Go modules

## Motivation and Context

This change moves the second OpenFaaS project to use Go modules,
but with the vendor folder retained. It is required as part of
the work to amalgamate faas-netes and the OpenFaaS Operator
which has already switched over to modules and the Go 1.17
client for Kubernetes.

## How Has This Been Tested?

Tested with a make in Docker and with the unit tests.


stefanprodan:
Please remove the vendor dir, go 1.13 makes use of the go proxy and downloading the deps is super fast.

alexellis:
Hi @stefanprodan, we're not moving away from the vendor folder just yet, I'd like to get all projects migrated to modules first, then at a later date we can evaluate whether we the vendor folder will be removed.

I did mention this on the review of the openfaas-operator, not sure if you saw that yet? https://github.com/openfaas-incubator/openfaas-operator/pull/104#pullrequestreview-344995443

Alex

Issue #586, opened 2020-01-25
[Feature] Profile support for taints, tolerations, anti/affinity and hardware


alexellis:
## Description

Add a profiles feature to give support for named sets of configuration for types of nodes or compute. Each profile will support for taints, tolerations, anti/affinity and hardware

## Use-case

* A common request is to add a toleration so that functions can be scheduled on spot instances, but we have so far been reluctant to leak that kind of complicated notation into the stack.yml file.

* TPU or GPU access - also a very common ask, this can be done with node-selectors, but node-selectors will be "going away" in favour of much more complex affinity rules

* Anti-affinity - imagine a critical function that should be scheduled across a unique set of nodes, Kubernetes has a answer for this.

* Security context - for instance when an IAM role is mounted by AWS and is unreadable by the function due to the fs group

* Runtimes - gVisor and kata containers require a "runtimeClass" to enable their runtime, for isolating workloads.

Feel free to add more use-cases below, if you've run into them in your line of business.

### Definition of a profile

Each named profile would be configured for the faas-netes container, not in stack.yml, it would have sections for each part of Kubernetes config needed and then allocated to any matching deployment request.

### Selecting a profile

A profile would be selected by adding an annotation in stack.yml such as `com.openfaas.profile: tpu`

## Influences

[Michael Crosby](https://github.com/crosbymichael?tab=repositories) mentioned the idea of profiles for CNI drivers which inspired this idea. I also socialised it with @stefanprodan and @LucasRoesler who seemed to prefer it to leaking abstractions into the stack.yml file.

I also discussed this use-case with the Knative team at KubeCon who stated that they didn't want to support Kubernetes scheduling rules - in so much that "that feels Serverful, not Serverless", and that low-priority / high-priority profiles may fit better.

martindekov:
Feel like this is interesting plus I haven't touched the backend of the project. Can I look into this if no one started working on it ?

alexellis:
Feel free to explore it and report back findings.

alexellis:
Maybe you can start with a hardcoded patch as an example of how we can enable the easiest usecase.

LucasRoesler:
This reminds me of how PVC and Volumes work with storage classes in k8s https://kubernetes.io/docs/concepts/storage/storage-classes/. This could be a nice parallel to this idea

alexellis:
Thanks for the comment, can you give a quick example @LucasRoesler ?

LucasRoesler:
[...]

PR #595, opened 2020-03-04
Exploratory: Enable volumes mounting in functions

martindekov:
Enabling volumes to be mounted in function's file
system using annotations

Signed-off-by: Martin Dekov <mdekov@vmware.com>

# Disclaimer

*This PR is exploratory and not solution discussed with the community.*

## Description

Functions can now mount volumes with this PR

## Motivation and Context

I needed the functionality and decided that this can be used by someone inspiration as part of the profiles support: #586 

Also was suggested in:
https://github.com/openfaas/faas/issues/1178
and:
https://github.com/openfaas/faas/issues/320

This POC does exactly that.

## How Has This Been Tested?

Image in:
`martindekov/faas-netes/0.0.2`

With function applying the following annotations:
```
...
    annotations:
      com.openfaas.volumename: "some-pv-storage"
      com.openfaas.pvcname: "some-claim"
      com.openfaas.pvcmountpath: "/some/directory"
...
```

The function's deployment had those fields:
```
...
        volumeMounts:
        - mountPath: /some/directory
          name: some-pv-storage
        - mountPath: /var/openfaas/secrets
          name: some-projected-secrets
          readOnly: true
      dnsPolicy: ClusterFirst
...
      volumes:
      - name: some-pv-storage
        persistentVolumeClaim:
          claimName: some-claim

...
```

And the volume was mounted successfully.

## Types of changes

- [x] POC not to be integrated or

## Checklist:

### DIFF
--- a/handlers/deploy.go
+++ b/handlers/deploy.go
@@ -142,11 +142,26 @@ func makeDeploymentSpec(request types.FunctionDeployment, existingSecrets map[st
 	annotations := buildAnnotations(request)
 
 	var serviceAccount string
+	var volumeName string
+	var pvcMountPath string
+	var pvcName string
 
 	if request.Annotations != nil {
 		annotations := *request.Annotations
 		if val, ok := annotations["com.openfaas.serviceaccount"]; ok && len(val) > 0 {
 			serviceAccount = val
+		}
+
+		if val, ok := annotations["com.openfaas.volumename"]; ok && len(val) > 0 {
+			volumeName = val
+		}
+
+		if val, ok := annotations["com.openfaas.pvcmountpath"]; ok && len(val) > 0 {
+			pvcMountPath = val
+		}
+
+		if val, ok := annotations["com.openfaas.pvcname"]; ok && len(val) > 0 {
+			pvcName = val
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


alexellis:
Just a thought here. If this was added, would it cover every use-case if only one volume was supported per function?

handlers/deploy.go
```diff
@@ -142,12 +142,27 @@ func makeDeploymentSpec(request types.FunctionDeployment, existingSecrets map[st
 	annotations := buildAnnotations(request)
 
 	var serviceAccount string
+	var volumeName string
```
alexellis: You'd have to handle updates too, if someone wanted to remove a volume that was already attached.

handlers/deploy.go
```diff
@@ -207,11 +222,27 @@ func makeDeploymentSpec(request types.FunctionDeployment, existingSecrets map[st
 							SecurityContext: &corev1.SecurityContext{
 								ReadOnlyRootFilesystem: &request.ReadOnlyRootFilesystem,
 							},
+							VolumeMounts: []apiv1.VolumeMount{
```
alexellis: This adds a volume even if one isn't configured, so you should set up the slice outside of this piece of code. ideally in a method to create it.

handlers/deploy.go
```diff
@@ -207,11 +222,27 @@ func makeDeploymentSpec(request types.FunctionDeployment, existingSecrets map[st
...
+									Name:      volumeName,
+									MountPath: pvcMountPath,
+								},
+							},
 						},
 					},
 					ServiceAccountName: serviceAccount,
 					RestartPolicy:      corev1.RestartPolicyAlways,
 					DNSPolicy:          corev1.DNSClusterFirst,
+					Volumes: []apiv1.Volume{
```
alexellis: Same as comment above

martindekov:
>Just a thought here. If this was added, would it cover every use-case if only one volume was supported per function?
Added this as POC if we want to merge it in master I can bring this in a better state.

martindekov:
I will clear the shortcomings. I am aware of those but was not sure whether we want this to enter master. I suppose we do want to integrate the change and I will update the PR and fix all issues you mentioned. For the question whether only one volume is supported per function - yes this is the main idea. We might be able to extend this with wildcard annotation. So the key of the annotation itself suggest the volume name path etc. But I believe this would not be needed, at least I don't need more than one volume mounted.

alexellis:
I think you should work to fix these basic issues, since you've gone ahead with a public PR, we can at least have it at a quality that anyone who lands here can test it or try it out. Would you be open to fixing up the initial feedback?

martindekov:
I will close this, but will continue refining it and will keep the change in the branch.

PR #600, opened 2020-03-24
Move CI to use helm3

alexellis:
Helm 3 removes the need for tiller. This makes the installation
faster and use less bandwidth.


alexellis:
There appears to be issues with Travis CI right now with its reporting, but the build did pass.

Issue #491, opened 2019-08-09
Merge faas-operator with faas-netes


stefanprodan:
Port operator's code and CRD to faas-netes and allow switching between the two controllers with a command flag. This will enable us to add new features to both controllers at the same time without code duplication.

Code merge phases:
1. Move the operator code as it is
2. Add `-mode=crd` command flag to start the crd operator (defaults to `-mode=` faas-netes mode)
3. Extract common code in the `k8s` package and trim down the code duplication between operator and faas-netes

alexellis:
This has been completed now. I've moved over the README and added a note on the operator repo.

PR #607, opened 2020-04-09
Merge the Operator with faas-netes

alexellis:
## Description

Merge the Operator with faas-netes

## Motivation and Context

- [x] I have raised an issue to propose this change ([required](https://github.com/openfaas/faas/blob/master/CONTRIBUTING.md)) #491 

This commit merges the code of the operator into the faas-netes
repo, so that the two controllers can be worked on together
and kept in sync.

To be done:
- [x] Copy in the operator code
- [x] Run codegen and build
- [x] Run faas-netes e2e tests during CI
- [x] Add operator e2e tests to CI

## How Has This Been Tested?

* The code builds
* Update codegen worked as expected
* The Dockerfile builds
* I deployed faas-netes and deployed figlet / invoked it
* I deployed the operator and deployed nodeinfo and invoked it

Please do try it out for yourself:

### For faas-netes (default)

```
arkade install openfaas \
  --set faasnetes.image=openfaas/faas-netes:0.10.2-rc2
```

To try it out with the operator:

```
arkade install openfaas \
 --set operator.create=true \
 --set operator.image=openfaas/faas-netes:0.10.2-rc2
```

Then use:

```
kubectl logs deploy/gateway -n openfaas  faas-netes -f
```

### Test the operator

Then update the command for the operator container to use the flag "-operator" which is false by default.

Patch the flag with something like:

```
kubectl patch -n openfaas deploy/gateway  \
-p='[{"op": "add", "path": "/spec/template/spec/containers/1/command", "value": ["./faas-netes", "-operator=true"]} ]' --type=json
```

Then:

```
kubectl logs deploy/gateway -n openfaas  operator -f
```

## Types of changes


alexellis:
Whilst testing the store with the operator, I found this issue: https://github.com/openfaas/faas-cli/issues/792

alexellis:
Just to clarify for anybody looking, the goal of this patch is not to optimise and deduplicate code but to lift and shift the operator into the faasnetes binary.

alexellis:
Here is the release tag which will have a corresponding Docker image in 10-20 minutes - https://github.com/openfaas/faas-netes/releases/tag/0.10.2

@stefanprodan @LucasRoesler I de-duplicated the version struct, which is relatively minor, but what other areas can we de-duplicate and consolidate now?

Alex

LucasRoesler:
@alexellis i think the factory methods and the CRUD actions to the standard k8s api objects should be standardized. 

The operator should use the same FunctionFactory etc. Ideally, we would create a manager/wrapper for the Function logic that can then be shared between the API handler and the operator's controller. 

The API endpoints can probably also be partially consolidated; specifically secrets, namespaces, info, invoke, and logs can just be the same implementation.  The endpoints that contains the substring "function" probably need to be different depending on the flag.

PR #620, opened 2020-04-23
Enable namespaces for logs

alexellis:
The namespace was hard-coded to the default, this updates the
code to make a logical choice based upon the user request.

Deletes "rouge" files I introduced by accident after demoing
codegen on the openfaas members call.


alexellis:
KinD appears to be having a bad hair day.

PR #627, opened 2020-05-01
Update helm chart to include max_inflight

alexellis:

## Description

Update helm chart to include max_inflight

## Motivation and Context

max_inflight was introduced in queue-worker 0.11.0 to enable
multiple concurrent inflight requests without scaling the
queue-worker deployment.

## How Has This Been Tested?

Tested by @matthiashanel and CI

## Types of changes

Part of https://github.com/openfaas/docs/issues/225


chart/openfaas/templates/queueworker-dep.yaml
```diff
@@ -60,6 +58,10 @@ spec:
         - name: faas_function_suffix
           value: ".{{ .Values.functionNamespace }}.svc.{{ .Values.kubernetesDNSDomain }}"
         {{- end }}
+        {{- if .Values.queueWorker.maxInflight }}
+        - name: max_inflight
+          value: ".{{ .Values.queueWorker.maxInflight }}"
```
LucasRoesler, line 63: Is this `.` prefix correct?
```suggestion
          value: "{{ .Values.queueWorker.maxInflight }}"
```

chart/openfaas/templates/queueworker-dep.yaml
```diff
@@ -60,6 +58,10 @@ spec:
         - name: faas_function_suffix
           value: ".{{ .Values.functionNamespace }}.svc.{{ .Values.kubernetesDNSDomain }}"
         {{- end }}
+        {{- if .Values.queueWorker.maxInflight }}
```
LucasRoesler, line 61: Given the implementation, this if statement isn't really needed, any integer value is "valid" and the queue worker will handle values <= 1 correctly, we might as well simplify the chart here by removing the if block

chart/openfaas/templates/queueworker-dep.yaml
alexellis: No that's a good catch thanks

chart/openfaas/templates/queueworker-dep.yaml
alexellis: I suspect it wouldn't handle 0 and < 1

alexellis:
@LucasRoesler I've forced pushed both of your suggested changes.

chart/openfaas/templates/queueworker-dep.yaml
LucasRoesler: It does infact handle that case, i checked it already

LucasRoesler: When it only starts the parallel goroutines when the value is greater than one, otherwise it is the default single handler https://github.com/openfaas/nats-queue-worker/pull/95/files#diff-e871404fd69cab40521d6b2178847cf7R77

matthiashanel: line 75 passes it to stan. Unfortunately that is not as forgiving. ... 
If it keeps the helm portion simpler I can move that line into the if...

matthiashanel: just let me know what you want me to do about that.

matthiashanel: @alexellis , since you merged I'll fix it in the queue worker

PR #631, opened 2020-05-07
Update apply REST handler

alexellis:
The HTTP handler appeared to generate errors with the latest
versions of Kubernetes (1.17.5 and 1.18). A function created
via REST and then updated via a CRD or visa-versa caused an
error about the RevisionVersion being required.

By getting the function object before updating it, this
appears to resolve the issue.

### DIFF
--- a/pkg/server/apply.go
+++ b/pkg/server/apply.go
@@ -2,16 +2,17 @@ package server
 
 import (
 	"encoding/json"
+	"fmt"
 	"io/ioutil"
 	"net/http"
-	"strings"
 
 	faasv1 "github.com/openfaas/faas-netes/pkg/apis/openfaas/v1"
 	clientset "github.com/openfaas/faas-netes/pkg/client/clientset/versioned"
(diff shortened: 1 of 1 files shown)


alexellis:
Wasn't keen to write fixes 630 since there's more to it.

LucasRoesler:
@alexellis can you provide some more details on what else needs to be covered/handled?  When i read the original issue, the only true error I saw was the one about ResourceVersion and this will clearly handle that case.

alexellis:
Hi @LucasRoesler the comment is about https://github.com/openfaas/faas-netes/issues/634 - I wanted to have it covered and tracked in a separate issue so that it doesn't get forgotten about. I'm happy to merge this fix when CI passes.

CI seems very flakey at the moment with KinD when it creates its second cluster for e2e tests.

Alex

alexellis:
A separate PR was opened by @LucasRoesler to do the same thing this morning in: #633

PR #648, opened 2020-06-09
[WIP] Add Ready condition to Function CustomResource when using Operator

brandonkal:
This adds a Ready condition to the Function resource status and an observedGeneration field.

## Motivation and Context

This is required in order to tell if a Function deployment has succeeded.

- [x] I have raised this issue #649  ([required](https://github.com/openfaas/faas/blob/master/CONTRIBUTING.md))

## How Has This Been Tested?

This has been verified on my local development cluster, and the status does change as expected.

## Types of changes

## Checklist:


pkg/apis/openfaas/v1/types.go
```diff
@@ -46,6 +46,20 @@ type FunctionResources struct {
 	CPU    string `json:"cpu,omitempty"`
 }
 
+// FooStatus is the status for a Foo resource
```
LucasRoesler, line 49: this comment seems wrong

pkg/apis/openfaas/v1/types.go
```diff
@@ -46,6 +46,20 @@ type FunctionResources struct {
...
+type FunctionStatus struct {
+	Conditions          []FunctionCondition `json:"conditions"`
+	UnavailableReplicas int32               `json:"unavailableReplicas"`
+	ObservedGeneration  int64               `json:"observedGeneration"`
+}
+
+const Ready = "Ready"
+
+type FunctionCondition struct {
```
LucasRoesler, line 58: this needs a comment

pkg/controller/controller.go
```diff
@@ -326,10 +326,43 @@ func (c *Controller) syncHandler(key string) error {
 		return err
 	}
 
+	// Finally, we update the status block of the Foo resource to reflect the
+	// current state of the world
+	err = c.updateFunctionStatus(function, deployment)
```
LucasRoesler, line 331: this update will not accurately track the status of the deployment. This loop will only run (if I recall correctly) when there is a change to a Function resource. But the change in replica state of a deployment (which is really a change on the ReplicaSet) will not trigger this change, So this status will generally not be accurate unless you have a listener for the Pod events and track those manually

brandonkal: The intent here is to watch the status during the rollout. So watching the Pod after deployment is out of scope as far as I can see (especially with scale-to-zero). If we want that, what would I change?

LucasRoesler: I am not sure this achieves that, have you verified it updates as the deployment pods reach Available?  you will need to use a function that has a  long startup time (10s>) to really very this.

The `syncHandler` only runs on items in the `workQueue`
https://github.com/openfaas/faas-netes/blob/b1db3a1f6c423c98904a529bd5d53bf5d3063f9d/pkg/controller/controller.go#L182

but workQueue only ever adds Functions via this method https://github.com/openfaas/faas-netes/blob/b1db3a1f6c423c98904a529bd5d53bf5d3063f9d/pkg/controller/controller.go#L376 
which is only called on function change https://github.com/openfaas/faas-netes/blob/b1db3a1f6c423c98904a529bd5d53bf5d3063f9d/pkg/controller/controller.go#L116-L120

This means you are only going to process the Deployment status when the Function object is changed, meaning the client has modified the spec in someway.  

This implementation wont see any Pod/Deployment related events, which you are trying to track.  For this to work you would need to create a pod/deploy informer that tracks pods events, when it sees a pod event check the annotations to decide if it is a function pod/deploy, and then uses that to update the status , it will look similar to this https://github.com/openfaas/faas-netes/blob/b1db3a1f6c423c98904a529bd5d53bf5d3063f9d/pkg/controller/controller.go#L116

With respect to scale-to-zero, I think it requires that we inspect the request replica count from the Deployment object, so that we can convey that the "Available" count is zero because it was requested to be zero

Finally, I don't think it is realistic to scope the "Status" to only during deployment. As a user, when I see `Function Status` it is not obvious that it would be a one time "when it was created/updated" status check. This is not how other objects in k8s behave and this would be surprising or a bug if it does not update.

alexellis:
> I have proposes this change on Slack (required)

@brandonkal you will need to go ahead and raise an issue to propose this change. See the way we take suggestions in the contributing guide -> https://github.com/openfaas/faas/blob/master/CONTRIBUTING.md#i-want-to-contribute-on-github

pkg/controller/controller.go
alexellis: > So watching the Pod after deployment is out of scope as far as I can see (especially with scale-to-zero)

This is why we ask all contributors to raise an issue for discussion and to get approval on design before they go on to write patches and get too attached to a particular solution.

If there is to be a ready status, then I believe Kubernetes developers would expect it to reflect true readiness, not whether an initial deployment worked or not.

brandonkal:
I plan to address the feedback here. As it is not clear what is wanted here now, it may be some time before I can fully address this.

pkg/apis/openfaas/v1/types.go
brandonkal: Addressed

pkg/controller/controller.go
brandonkal: Just tested the code with longer-running deployments. It works but the Function.Status is not updated until the next reconcile loop. Oddly this was 2 minutes for my case (the code suggests 5 minutes if I'm reading that correctly). I'll look to watch deployments. Thank you @LucasRoesler

brandonkal:
It's working now. Ready for another review @LucasRoesler

pkg/controller/controller.go
```diff
@@ -325,11 +345,55 @@ func (c *Controller) syncHandler(key string) error {
...
+				Type:   "Ready",
+			},
+		}
+		fnCopy.Status.UnavailableReplicas = deployment.Status.UnavailableReplicas
+		fnCopy.Status.ObservedGeneration = fn.ObjectMeta.Generation
+		// If the CustomResourceSubresources feature gate is not enabled,
+		// we must use Update instead of UpdateStatus to update the Status block of the Function resource.
+		// UpdateStatus will not allow changes to the Spec of the resource,
+		// which is ideal for ensuring nothing other than resource status has been updated.
+		_, err = c.faasclientset.OpenfaasV1().Functions(fn.Namespace).UpdateStatus(fnCopy)
```
LucasRoesler, line 391: One improvement that will make this easier to test and maintain is to split this into two methods, one that calculates the new stats and one that does the update. That will allow us to write a simple unit test that verifies how  the status is calculated. Having the tests that show how specific cases change (or dont) will make it much easier to maintain

alexellis:
Closing inactive PRs. If anyone needs this feel free to create a new issue.

PR #632, opened 2020-05-07
WIP Update go dependencies

aledbf:
## Description

- Go 1.13.9 is required because of `errors.Is` in API machinery
- Use [clientgofix](https://github.com/kubernetes-sigs/clientgofix) to update client-go calls

## Motivation and Context

- [ ] I have raised an issue to propose this change ([required](https://github.com/openfaas/faas/blob/master/CONTRIBUTING.md))

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [ ] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] I've read the [CONTRIBUTION](https://github.com/openfaas/faas/blob/master/CONTRIBUTING.md) guide
- [ ] I have signed-off my commits with `git commit -s`
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.

PR #654, opened 2020-06-15
Generate CRD YAML from code and add Policy resource

alexellis:
## Description

Generate CRD YAML from code and add Policy resource

## Motivation and Context

Part of the profiles work #586 

This introduces the Policy resource, but also adds `./hack/update-crds.sh` to generate the YAML for the CRD from the code. This means less maintenance over time and with the complexity of the Policy fields (affinity etc) makes this task easier overall.

`controller-gen` is part of the kubebuilder project.

* Regenerates CRDs and adds Policy type
* Updates client-go to K8s 1.18
* Updates files for new changes using clientgofix which adds
context.TODO() where required. At a later date this needs to
be changed to any context we happen to be using in the
function.

## How Has This Been Tested?

Not tested **yet**

## Types of changes

This could be a breaking change, but in theory should not be - if the generated CRD works as well as the manually created one.

## Checklist:

Thanks to @munnerz for helping out with suggestions on K8s Slack


go.mod
```diff
@@ -9,19 +9,17 @@ require (
...
+	github.com/prometheus/client_golang v1.0.0
 	go.uber.org/goleak v1.0.0 // indirect
-	golang.org/x/sys v0.0.0-20190920190810-ef0ce1748380 // indirect
 	google.golang.org/appengine v1.6.2 // indirect
 	gopkg.in/check.v1 v1.0.0-20190902080502-41f04d3bba15 // indirect
-	k8s.io/api v0.17.4
-	k8s.io/apimachinery v0.17.4
-	k8s.io/client-go v0.17.4
-	k8s.io/code-generator v0.17.4
+	k8s.io/api v0.18.2
```
alexellis, line 19: The change updated the Go client - we will need to check if this breaks any compatibility or removes the ability to use older versions of k8s.

chart/openfaas/templates/crd.yaml
```diff
@@ -65,5 +65,27 @@ spec:
...
+    - name: v1alpha1
+      served: true
+      storage: false
+  names:
+    plural: policies
+    singular: policy
+    kind: Policy
+    shortNames:
+    - pol
+  scope: Namespaced
```
munnerz, line 89: What's the strategy for copying/patching the generated schema into your CRD templates for the Helm chart?

For reference, we achieve this in cert-manager by adding in Helm templating directives dynamically (i.e. at the time the Helm chart is built), by appending and prepending the required `{{ if .. }}` and `{{ end }}` blocks.

main.go
```diff
@@ -32,6 +32,8 @@ import (
 
 	// required for updating and validating the CRD clientset
 	_ "k8s.io/code-generator/cmd/client-gen/generators"
+	// main.go:36:2: import "sigs.k8s.io/controller-tools/cmd/controller-gen" is a program, not an importable package
```
munnerz, line 35: More info on how to avoid this: https://github.com/go-modules-by-example/index/blob/master/010_tools/README.md#tools-as-dependencies

go.mod
munnerz: > or removes the ability to use older versions of k8s.

cert-manager continues to support k8s all the way back to version 1.11 using the 1.18 dependencies: https://testgrid.k8s.io/jetstack-cert-manager-master#ci-cert-manager-e2e-v1-11

It will mean you need to change your calls to the kubernetes client interfaces to include `context.Context` as the first argument - liggitt put together a tool for automating this if you have a lot of call-sites.

alexellis:
Related to #632 cc @aledbf

chart/openfaas/templates/crd.yaml
alexellis: I was going to copy and paste it into the middle of the helm chart file after each change -> https://github.com/openfaas/faas-netes/blob/master/chart/openfaas/templates/crd.yaml

We could update the bash script to populate this file with a prefix/suffix and the generated content in the middle.

main.go
alexellis: I was just wondering what the advantage would be of compiling this into the code vs doing `go get` inside the update-crd.sh script?

munnerz: Thanks to the `// build` tag, this dependency will not be built into your binaries 😄  this is just used to have go modules track the dependency for you as part of your module.

`go get` will also work, although you'll need to be careful to `export GO111MODULE=off` before to avoid it modifying your actual project's `go.mod` file.

alexellis: I've tried this and cannot seem to get `go test ./...` to be happy, even with `RUN go test -v $(go list ./... | grep -v /tools/)` 🤔

Issue #616, opened 2020-04-17
[Feature] Operator - add support for managing multiple-namespaces


alexellis:
## Expected Behaviour

Add support for the namespaces endpoint, as per faas-netes

## Current Behaviour

Currently unavailable (404)

## Possible Solution

Quick fix / work-around: implement a stub that returns the default namespace

Longer term fix: copy patches from faas-netes, test e2e with the operator

cc @stefanprodan @LucasRoesler

In order to have this working e2e, we may also need (via Bruno):

Make the operator watch resources across all Kubernetes namespaces.
Introduce the required RBAC policies (cluster role and cluster role binding).
Copy whatever faas-netes is doing as per #511

andrevtg:
Quick suggestion: while scanning multiple namespaces can be useful, it should be possible to configure the operator to scan a limited and known set of namespaces, or the operator will break in more constrained environments.

alexellis:
We have a cluster-role available now for the operator in Helm, to allow management of multiple namespaces.

andrevtg:
Can we make "ClusterRole" optional? When deploying into namespace restricted clusters we cannot use it and we also should limit which namespaces to watch.

alexellis:
There was a question from a contributor twice on this pull request. Both times I got the email and then he deleted it immediately afterwards.

In any case he was asking whether this would change the behaviour of the operator to only work as a ClusterRole. The answer is no, it just adds it as an option for those who want to use multiple namespaces with a single installation of OpenFaaS.

andrevtg:
It was me :)  I saw the chart option and then deleted the question, sorry

PR #668, opened 2020-07-04
Add namespaces support for REST

martindekov:
Adding namespaces support for the REST endpoints
of the operator. Currently only the function objects
are spawned without the actual pods running the function
themselves

Signed-off-by: Martin Dekov <mvdekov@gmail.com>

Add function CRD creation across namespaces for the operator's REST controller

## Description

Enable rest controller for the operator to spawn function objects across namespaces. The functions yet don't spawn pods, only the function CRD.

## Motivation and Context

Part of #616 

## How Has This Been Tested?

Tested by creating resource, will publish logs once we discuss this draft

## Types of changes

## Checklist:

### DIFF
--- a/chart/openfaas/values.yaml
+++ b/chart/openfaas/values.yaml
@@ -73,7 +73,7 @@ oauth2Plugin:
   securityContext: true
 
 faasnetes:
-  image: openfaas/faas-netes:0.10.5
+  image: martindekov/faas-netes:0.12.0-rc1
   readTimeout : "60s"
   writeTimeout : "60s"
   imagePullPolicy : "Always"    # Image pull policy for deployed functions
--- a/pkg/server/delete.go
+++ b/pkg/server/delete.go
@@ -13,8 +13,21 @@ import (
 	glog "k8s.io/klog"
 )
 
-func makeDeleteHandler(namespace string, client clientset.Interface) http.HandlerFunc {
+func makeDeleteHandler(defaultNamespace string, client clientset.Interface) http.HandlerFunc {
 	return func(w http.ResponseWriter, r *http.Request) {
+
+		q := r.URL.Query()
+		namespace := q.Get("namespace")
+
+		lookupNamespace := defaultNamespace
+		if len(namespace) > 0 {
+			lookupNamespace = namespace
+		}
+
+		if namespace == "kube-system" {
+			http.Error(w, "unable to list within the kube-system namespace", http.StatusUnauthorized)
+			return
+		}
 
 		if r.Body != nil {
 			defer r.Body.Close()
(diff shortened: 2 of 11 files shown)


chart/openfaas/values.yaml
```diff
@@ -73,7 +73,7 @@ oauth2Plugin:
   securityContext: true
 
 faasnetes:
-  image: openfaas/faas-netes:0.10.5
+  image: martindekov/faas-netes:0.12.0-rc1
```
martindekov, line 76: I will revert, generally the change is in this image

pkg/server/secret.go
```diff
@@ -23,159 +16,9 @@ const (
...
-	if err != nil {
-		return secrets, err
-	}
-
-	for _, item := range res.Items {
-		secret := faastypes.Secret{
-			Name: item.Name,
-		}
-		secrets = append(secrets, secret)
+	handler := handlers.SecretsHandler{
```
LucasRoesler, line 19: It isn't clear to me why we make `SecretsHandler` public instead of using the already public factory method `MakeSecretHandler`

pkg/server/delete.go
```diff
@@ -13,9 +13,22 @@ import (
 	glog "k8s.io/klog"
 )
 
-func makeDeleteHandler(namespace string, client clientset.Interface) http.HandlerFunc {
+func makeDeleteHandler(defaultNamespace string, client clientset.Interface) http.HandlerFunc {
```
LucasRoesler, line 16: I am going to make a rather big suggestion that will require another refactor.  I think we should have a `FunctionClient` interface with the needed create/update/delete/list methods. We can then create two implementation one based on the clientset and one based on the informer/lister and even one based on the CRD. This allows the logic in the handlers to be separated from the implementation.  This will simplify the handlers and allow us to easily swap the implementation over time. This also means the server implementation can be unified and reused between both faas-netes and the operator because the important distinct "CRD based FunctionClient" vs "API based FunctionClient" is wrapped into the specific client implementation

pkg/server/secret.go
martindekov: Yeah Lucas, now I can see that due to your feedback, when I wrote this I didn't thought that actually secrets are generic openfaas feature, and not specific for the operator. Will switch to that :+1:

pkg/server/delete.go
martindekov: From my understanding on the proposal, you mean using only one `types.FaaSHandlers` instead of  having almost duplicate code for the operator and faas-netes. This `FunctionClient` will internally switch between the details of the two (if there are any differences between them). So essentially is this some sort of a Facade pattern over the types.FaaSHandlers? I can ping you to discuss this further to gather more context on the proposed refactoring. Also this change is missing the actual creation of functions, it currently creates only the Function CRD. I was also hoping to get some insight what I should be changing in order to make this work e2e.

martindekov:
Full change is here: https://github.com/openfaas/faas-netes/pull/673

PR #671, opened 2020-07-20
Move misc packages into pkg package

alexellis:

## Description

Move misc packages into pkg package

## Motivation and Context

This is one of the main bits of feedback from @StefanProdan
that was still remaining after consolidating the two
controllers.

This patch also removes the service namespace field from
the scale type. The value can and is read from the query
string, meaning that this field was redundant and non-standard
compared with the types in faas-provider.

## How Has This Been Tested?

Tested with go test ./... and with make using the Dockerfile.

## Types of changes

Refactoring, no changes unless someone is depending on one of the public packages, in which case they can continue to use an older version.


pkg/handlers/replicas.go
```diff
@@ -48,10 +48,6 @@ func MakeReplicaUpdater(defaultNamespace string, clientset *kubernetes.Clientset
 			}
 		}
 
-		if len(req.ServiceNamespace) > 0 {
```
Waterdrips, line 51: This looks unrelated to the renames. Was this change expected in this pr?

LucasRoesler: It is related in so far as not needing to temporarily create a `pkg/types/requests.go`

alexellis: This is in the commit message:

```
This patch also removes the service namespace field from
the scale type. The value can and is read from the query
string, meaning that this field was redundant and non-standard
compared with the types in faas-provider.
```

alexellis: Not an issue on this occasion, but thanks for the keen eye, this is why reviews are useful 👍

PR #683, opened 2020-08-29
Support function resolution in multiple namespaces

LucasRoesler:
## Description

**What**
- Ensure that the pass only the function name to the Endpoint lister,
  because the namespace is already set when constructing the lister.
  The existing code would always return a 404 because it sent the
  "function_name.namespace" as the function name. This can never be
  found since it is not a valid service name
- Update the error messages to make the component parts of the message 
   easier to read.  It will now be easier to see which part is the error from the 
   k8s api.

## Motivation and Context

Resolves openfaas/faas#1565

## How Has This Been Tested?

The unit test for the proxy resolver has been updated to catch this case and it now passes.

It has also been tested locally using 

```sh
kind create cluster

# doing this will cause extras and openfaas-fn to show in the UI
kubectl create ns extras
kubectl annotate ns openfaas-fn  openfaas=1
kubectl annotate ns extras  openfaas=1

arkade install openfaas --basic-auth=false --direct-functions=false --operator --clusterrole --set operator.image=theaxer/faas-netes:6802054

# in a separate terminal
kubectl port-forward svc/gateway 8080
```

open http://localhost:8080 and use the UI to deploy and invoke the function 

image

## Types of changes

## Checklist:

### DIFF
--- a/pkg/k8s/proxy.go
+++ b/pkg/k8s/proxy.go
@@ -55,10 +55,14 @@ func getNamespace(name, defaultNamespace string) string {
 }
 
 func (l *FunctionLookup) Resolve(name string) (url.URL, error) {
-
+	functionName := name
 	namespace := getNamespace(name, l.DefaultNamespace)
 	if err := l.verifyNamespace(namespace); err != nil {
 		return url.URL{}, err
+	}
+
+	if strings.Contains(name, ".") {
+		functionName = strings.TrimSuffix(name, "."+namespace)
 	}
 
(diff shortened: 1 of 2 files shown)


Amalkh5:
@LucasRoesler I've tested these changes and it working as expected. 

Screenshot 2020-08-31 at 11 17 39 pm

LucasRoesler:
Thanks 💪

alexellis:
Thank you @amalkh5

alexellis (approved):
Approved

Issue #686, opened 2020-09-19
Profile CRD being created on every helm chart installation


SpaWn2KiLl:
`Profile` is being installed on every helm chart installation and there isn't any flag that prevents this behavior. Since a CRD is not namespace scoped, this breaks my development environment where every develop has his own sandbox with his own openfaas installation.

The first chart installation creates the kubernetes CRD but on the second installation it fails to install because the CRD is already installed.

## Expected Behaviour
Should be possible to install multiple openfaas charts on the same kubernetes cluster.

## Current Behaviour
On the second openfaas chart installation it prompts the following error:
```console
$ helm install openfaas openfaas/openfaas -n ns-a9d8a4 --version 6.0.1 --set basic_auth="true" --set serviceType="ClusterIP" --set faasIdler.dryRun="true" --set faasIdler.inactivityDuration="5m" --set faasIdler.reconcileInterval="5m" --set operator.create=true --set operator.createCRD=false --set istio.mtls=false --set functionNamespace="ns-a9d8a4"
Error: rendered manifests contain a resource that already exists. Unable to continue with install: CustomResourceDefinition "profiles.openfaas.com" in namespace "" exists and cannot be imported into the current release: invalid ownership metadata; annotation validation error: key "meta.helm.sh/release-name" must equal "openfaas": current value is "openfaas-dev"; annotation validation error: key "meta.helm.sh/release-namespace" must equal "ns-a9d8a4": current value is "openfaas-dev"
```

## Possible Solution
`operator.createCRD` should be used as a flag to install the CRD or not.

## Steps to Reproduce (for bugs)
1. Install an openfaas chart on a kubernetes cluster
  -
      ```console
      helm install openfaas openfaas/openfaas -n openfaas-dev --version 6.0.1 --set basic_auth="true" --set serviceType="ClusterIP" --set faasIdler.dryRun="true" --set faasIdler.inactivityDuration="5m" --set faasIdler.reconcileInterval="5m" --set operator.create=true --set operator.createCRD=true --set istio.mtls=false --set functionNamespace="openfaas-fn-dev"
      ```
2. Install a second openfaas chart on the same kubernetes cluster
  -
      ```console
[...]

PR #696, opened 2020-11-03
Profiles CRD toggle

alexellis:

## Description

Profiles CRD toggle

## Motivation and Context

Introduces a toggle for creating the Profiles CRD. This is
to help users overcome #686

The main use-case is for users who are installing within a 
restrictive environment, where they cannot create CRDs, but 
rely on an administrator to take on this role.

Makes a breaking change where "createCRDs" controls whether
any CRDs are created. This is on by default. Existing users
must take note when upgrading.

## How Has This Been Tested?

Perhaps some users would like to test?

createCRDs: false gave no CRDs in the cluster

createCRDs: true gave: profiles.openfaas.com

createCRDs: true with operator.create: true gave functions.openfaas.com and profiles.openfaas.com

createCRDs has been made "true" in the values.yaml file


alexellis:
This solution is similar to what was asked for here -> https://github.com/openfaas/faas-netes/pull/579#issuecomment-719593185 by @SpaWn2KiLl

alexellis:
@Gottox @SpaWn2KiLl @ArcticXWolf @RichterBIT

ArcticXWolf:
Shouldn't these RBAC definitions be disabled too, if profiles are not enabled?

https://github.com/openfaas/faas-netes/blob/08ffc578d054f407458a6279c31216d6155ef7da/chart/openfaas/templates/controller-rbac.yaml#L76
https://github.com/openfaas/faas-netes/blob/08ffc578d054f407458a6279c31216d6155ef7da/chart/openfaas/templates/controller-rbac.yaml#L185
https://github.com/openfaas/faas-netes/blob/08ffc578d054f407458a6279c31216d6155ef7da/chart/openfaas/templates/controller-rbac.yaml#L206

alexellis:
Potentially, however it is not causing an error at deployment time. Feel free to try it out.

ArcticXWolf:
I did, but in my cluster I do not have the rights to grant the `{{ .Release.Name }}-profiles` role (yes, I am currently in discussion with getting more access). However since this role is not needed, we dont need to deploy it? 

When I use the `{{- if .Values.createCRDs }}` in line 76 and 186 of controller-rbac.yaml I can deploy the chart fine.

alexellis:
That makes me think of something else. If we put those parts behind the same if statement, then what is the story for your administrator? How will he/she install those parts on your behalf?

ArcticXWolf:
Sorry, I am not quite sure what these CRDs are used for in OpenFaaS. Are they mandatory for the installation? I would've just skipped it for now (the chart seems to install just fine without) and instructed my administrator to either provide me with enough permissions at a later time or install the CRDs manually.

LucasRoesler (changes requested):
This adds a new value, which should be documented in the chart readme too https://github.com/openfaas/faas-netes/tree/master/chart/openfaas#configuration

alexellis:
I'll get that done @LucasRoesler - anything else, or otherwise happy for this to be merged?

LucasRoesler:
> I'll get that done @LucasRoesler - anything else, or otherwise happy for this to be merged?

That is it

alexellis:
OK, I've just done that.

LucasRoesler (approved):

ArcticXWolf:
So we keep the roles in the chart? Or shall this be addressed in a different issue?

alexellis:
Please can you raise a separate issue for discussion?

ArcticXWolf:
Sure thing, I've created #697 . Thanks for all your work 👍

alexellis:
I think it's been mentioned before but we can do even more to help customers through our [Premium Subscription](https://www.openfaas.com/support/) - take a look at it with your manager? Or [at our new GitHub Sponsors program](https://github.com/sponsors/openfaas/)

End-user companies are also invited to "chop wood and carry water" - we have a huge amount of unpaid work ahead of us moving all builds off Travis and the Docker Hub, both of which are no-longer free for OSS.

ArcticXWolf:
Sure, I'll ask 👍

Issue #717, opened 2020-11-27
Helm Chart creates ClusterRole even if clusterRole=false


dplusic:
## Expected Behaviour

Helm Chart should not create `ClusterRole` when `clusterRole=false`.

## Current Behaviour

It creates `ClusterRole` even if `clusterRole=false`.

## Possible Solution

It should create `Role` instead of `ClusterRole`.

## Steps to Reproduce (for bugs)

1. Prepare a service account without permission creating `ClusterRole`
2. Run `helm install openfaas openfaas/openfaas --set clusterRole=false --set createCRDs=false`
3. Error message: `Error: rendered manifests contain a resource that already exists. Unable to continue with install: could not get information about the resource: clusterroles.rbac.authorization.k8s.io "openfaas-prometheus" is forbidden: User "system:serviceaccount:default:default" cannot get resource "clusterroles" in API group "rbac.authorization.k8s.io" at the cluster scope`

## Context

I want namespaced installation of openfaas.

## Your Environment

* FaaS-CLI version ( Full output from: `faas-cli version` ): -

* Docker version `docker version` (e.g. Docker 17.0.05 ): -

* What version and distriubtion of Kubernetes are you using? `kubectl version` 
```
Client Version: version.Info{Major:"1", Minor:"19", GitVersion:"v1.19.4", GitCommit:"d360454c9bcd1634cf4cc52d1867af5491dc9c5f", GitTreeState:"clean", BuildDate:"2020-11-11T13:17:17Z", GoVersion:"go1.15.2", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"18+", GitVersion:"v1.18.9-eks-d1db3c", GitCommit:"d1db3c46e55f95d6a7d3e5578689371318f95ff9", GitTreeState:"clean", BuildDate:"2020-10-20T22:18:07Z", GoVersion:"go1.13.15", Compiler:"gc", Platform:"linux/amd64"}
```

* Operating System and version (e.g. Linux, Windows, MacOS): `Ubuntu 18.04 on WSL`

* Link to your project or a code example to reproduce issue: -

* What network driver are you using and what CIDR? i.e. Weave net / Flannel: -

alexellis:
This is expected, Prometheus needs to scrape in more than one namespace. The cluster role in question is for the OpenFaaS Operator.

Happy for you to send a PR, otherwise you'll need to wait for this to be prioritized.

Premium Subscription customers get priority.

[...]

PR #720, opened 2020-11-29
Limit the Prometheus role to two namespaces by default

alexellis:

## Description

Limit the Prometheus role to two namespaces by default

## Motivation and Context

Limit the Prometheus role to two namespaces

Unless clusterRole is specified, the Prometheus role will be
restricted to scraping from only a single namespace.

This fixes issue: #717 where I user complained that they did 
not want to create a ClusterRole in their cluster.

## How Has This Been Tested?

It has been tested with k3d and K8s 1.19 with and without the 
--set clusterRole=true flag passed into the faas-netes helm 
chart.

The second Role and RoleBinding needed a different name to the 
ones in the primary namespace in order for the RBAC error to 
go away in Prometheus.

## Types of changes

## Checklist:

This is likely to break for any users wishing to upgrade, due to the existing `ClusterRole` which they 
will have in their cluster and its mapping.

The upgrade procedure is to delete the roles and bindings and then to run a helm upgrade.

### DIFF
--- a/chart/openfaas/templates/prometheus-cfg.yaml
+++ b/chart/openfaas/templates/prometheus-cfg.yaml
@@ -37,7 +37,9 @@ data:
             namespaces:
               names:
                 - {{ .Release.Namespace }}
-                - {{  $functionNs }}
+{{- if ne $functionNs (.Release.Namespace | toString) }}
+                - {{ $functionNs }}
+{{- end }}
         relabel_configs:
         - action: labelmap
           regex: __meta_kubernetes_pod_label_(.+)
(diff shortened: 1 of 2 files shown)


alexellis:
Thank you to @meyskens for suggesting that two Roles may be necessary. I tried two roles and bindings with the same name (but differing namespaces) which failed, but with unique names, it worked as expected.

alexellis:
@LucasRoesler just wondering when you could get to this?

LucasRoesler (approved):
Looks good and is working for me

alexellis:
Thank you

Issue #724, opened 2020-12-05
List namespaces warning


alexellis:
## Expected Behaviour

No errors about listing namespaces in the default, namespace-scoped mode.

## Current Behaviour

We get this error continually, as the gateway tries to list namespaces to collect metrics for each.

```
2020/12/05 10:24:29 Query namespaces
2020/12/05 10:24:29 Error listing namespaces: namespaces is forbidden: User "system:serviceaccount:openfaas:openfaas-controller" cannot list resource "namespaces" in API group "" at the cluster scope
```

## Possible Solution

We already have a cluster_scope variable. When it's set to false, we should just return the default namespace, and not try to list them.

## Steps to Reproduce (for bugs)

1. `arkade install openfaas`
2. `kubectl logs deploy/gateway -n openfaas faas-netes --tail 100 -f`

## Context

This is spamming the logs and confusing.

PR #725, opened 2020-12-05
Silence errors for namespace listing

alexellis:
Fixes: #724

## Description

Silence errors for namespace listing

## Motivation and Context

Fixes: #724

## How Has This Been Tested?

To be tested through CI

## Types of changes

### DIFF
--- a/pkg/handlers/namespaces.go
+++ b/pkg/handlers/namespaces.go
@@ -19,26 +19,29 @@ import (
 )
 
 // MakeNamespacesLister builds a list of namespaces with an "openfaas" tag, or the default name
-func MakeNamespacesLister(defaultNamespace string, clientset kubernetes.Interface) http.HandlerFunc {
+func MakeNamespacesLister(defaultNamespace string, clusterRole bool, clientset kubernetes.Interface) http.HandlerFunc {
 	return func(w http.ResponseWriter, r *http.Request) {
-		log.Println("Query namespaces")
+		// log.Println("List namespaces")
 
 		if r.Body != nil {
 			defer r.Body.Close()
 		}
 
-		res := ListNamespaces(defaultNamespace, clientset)
+		namespaces := []string{}
+		if clusterRole {
+			namespaces = ListNamespaces(defaultNamespace, clientset)
+		} else {
+			namespaces = append(namespaces, defaultNamespace)
+		}
 
-		out, err := json.Marshal(res)
+		out, err := json.Marshal(namespaces)
 		if err != nil {
 			glog.Errorf("Failed to list namespaces: %s", err.Error())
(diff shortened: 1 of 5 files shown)


pkg/handlers/namespaces.go
```diff
@@ -19,26 +19,29 @@ import (
...
 			defer r.Body.Close()
 		}
 
-		res := ListNamespaces(defaultNamespace, clientset)
+		namespaces := []string{}
+		if clusterRole {
+			namespaces = ListNamespaces(defaultNamespace, clientset)
+		} else {
+			namespaces = append(namespaces, defaultNamespace)
+		}
```
LucasRoesler, line 35: We could slightly simplify this using
```suggestion
		namespaces := []string{defaultNamespace}
		if clusterRole {
			namespaces = ListNamespaces(defaultNamespace, clientset)
		}
```

pkg/handlers/namespaces.go
```diff
@@ -19,26 +19,29 @@ import (
 )
 
 // MakeNamespacesLister builds a list of namespaces with an "openfaas" tag, or the default name
-func MakeNamespacesLister(defaultNamespace string, clientset kubernetes.Interface) http.HandlerFunc {
+func MakeNamespacesLister(defaultNamespace string, clusterRole bool, clientset kubernetes.Interface) http.HandlerFunc {
 	return func(w http.ResponseWriter, r *http.Request) {
-		log.Println("Query namespaces")
+		// log.Println("List namespaces")
```
LucasRoesler, line 24: Just delete it
```suggestion
```

pkg/handlers/replicas.go
```diff
@@ -102,6 +103,7 @@ func MakeReplicaReader(defaultNamespace string, clientset *kubernetes.Clientset)
 			lookupNamespace = namespace
 		}
 
+		s := time.Now()
```
LucasRoesler, line 106: This might be more valuable as an actual prometheus metric. Having it in the logs is nice, but it means a person is going to make judgments on a few values as opposed to being able to collect and aggregate real data into graphs etc.

LucasRoesler (changes requested):

pkg/handlers/namespaces.go
alexellis: Very open to a PR.

pkg/handlers/namespaces.go
alexellis: Agreed. I am still undecided about whether we should delete all happy-path logging and focus on the Prometheus metrics gathered from the gateway.

pkg/handlers/replicas.go
alexellis: I agree. I think it is in there already 👍 This is another example of a log message we should probably delete.

PR #726, opened 2020-12-09
Use an informer for getting replicas

alexellis:

## Description

Use an informer for getting replicas and functions

## Motivation and Context

Use an informer for getting replicas and functions

This should reduce rate-limiting on the Kubernetes API
server when faas-netes is making many calls. It changes
both the function list handler and the replica status
API.

The API returns much faster than before, so creates more HTTP
calls between the gateway and faas-netes, this value will
need to be tuned in the gateway to be slower between retries.

## How Has This Been Tested?

Tested with k3d and it worked as expected.
## Types of changes

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,7 +2,7 @@
 TAG?=latest
 export DOCKER_CLI_EXPERIMENTAL=enabled
 
-all: build-local
+all: build-docker
 
 local:
 	CGO_ENABLED=0 GOOS=linux go build -a -installsuffix cgo -o faas-netes
--- a/main.go
+++ b/main.go
@@ -155,6 +155,9 @@ func runController(setup serverSetup) {
 
 	endpointsInformer := kubeInformerFactory.Core().V1().Endpoints()
 
+	deploymentLister := kubeInformerFactory.Apps().V1().
+		Deployments().Lister()
+
 	log.Println("Waiting for openfaas CRD cache sync")
 	faasInformerFactory.WaitForCacheSync(stopCh)
 	setup.profileInformerFactory.WaitForCacheSync(stopCh)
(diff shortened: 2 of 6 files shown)


alexellis:
@stefanprodan what are your thoughts on this? Can you take a brief look and confirm whether this could be related to the interval / timing on the informer?

alexellis:
I see a note that says:

```
	// the sync interval does not affect the scale to/from zero feature
	// auto-scaling is does via the HTTP API that acts on the deployment Spec.Replicas
	defaultResync := time.Minute * 5
```

Usually scaling figlet from zero would be no more than 2-5s.

alexellis:
For some unknown reason, my k3d cluster is taking a very long time to scale up. This may be a red herring. I just saw `8.606626s`, then `13.206552s` with the original code.

alexellis:
Test with: `arkade install openfaas --set faasnetes.image=alexellis2/openfaas-faas-netes:inform-replicas --set gateway.image=openfaas/gateway:0.20.2`

alexellis:
Testing with tuning:

```
arkade install openfaas --set faasnetes.image=alexellis2/openfaas-faas-netes:inform-replicas --set gateway.image=openfaas/gateway:0.20.2 --set gateway.direct_functions=false --set faasnetes.imagePullPolicy=IfNotPresent --set openfaasImagePullPolicy=IfNotPresent
```

Resulted in: `2.485267s`

Overall this change seems to work as expected, but the gateway should now poll at a slower interval.

alexellis:
@LucasRoesler This works for the list functions call and individual function status call, the additional latency is a red herring. It also feels faster when invoking functions.

alexellis:
I'll be squashing the commits here and changing the commit message before merging into the codebase.

LucasRoesler (approved):
Code looks good, but i haven't had a chance to run it. If the certifier says 👍 then i trust it

alexellis:
Appreciated @LucasRoesler :+1:

alexellis:
The PR build passed, the branch build failed - due to a timing issue from what I can see in the blogs. This is consistent with what we've seen on the repo so far.

Issue #734, opened 2021-01-07
Suggestion: add NATS Streaming maximum payload option to helm chart


CarbonC:
Sorry I took that long, here is the completed template:

It is impossible to modify NATS streaming maximum payload from the chart. In some use cases you might want to put it to more than the default 1Mb

## Expected Behaviour
You should be able to pass a command line argument to increase the limit of NATS streaming payload

## Current Behaviour
Being able to pass a value different than the default one of 1mb would allow nats to accept bigger chunks of data and forward them to openfaas functions correctly

## Possible Solution
I implemented a solution that works very well on my stack on #735 

## Steps to Reproduce (for bugs)
If you want to see why its limiting, just increase nginx default payload limit, then send a payload of more than 1mb to an openfaas async function

## Context
I'm using openfaas functions to treat some chunks of IOT data, that are mostly around a few hundred kb but some are a bit bigger, up to 5mb maybe.

## Your Environment
* FaaS-CLI version ( Full output from: `faas-cli version` ):
CLI:
 commit:  598336a0cad38a79d5466e6a3a9aebab4fc61ba9
 version: 0.12.21

Gateway
 uri:     https://openfaas.getmoona.com
 version: 0.20.2
 sha:     9bbb25e3c7c4cc5cd355edb3a122f8c7812e32db
 commit:  Fix a bug that caused the services list to keep growing

Provider
 name:          faas-netes
 orchestration: kubernetes
 version:       0.12.12
 sha:           43f64cf9b30fd5f0f56514dee6b9fe01cfbfed45

* What version and distriubtion of Kubernetes are you using? `kubectl version`
Client Version: version.Info{Major:"1", Minor:"19", GitVersion:"v1.19.3", GitCommit:"1e11e4a2108024935ecfcb2912226cedeafd99df", GitTreeState:"clean", BuildDate:"2020-10-14T12:50:19Z", GoVersion:"go1.15.2", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"19", GitVersion:"v1.19.3", GitCommit:"1e11e4a2108024935ecfcb2912226cedeafd99df", GitTreeState:"clean", BuildDate:"2020-10-14T12:41:49Z", GoVersion:"go1.15.2", Compiler:"gc", Platform:"linux/amd64"}

* Operating System and version (e.g. Linux, Windows, MacOS):
Ubuntu 20.04

* Link to your project or a code 
* example to reproduce issue:
[...]

PR #735, opened 2021-01-07
Add NATS Streaming maximum payload option to helm chart

CarbonC:
Add NATS Streaming maximum payload option to helm chart

## Description
Add NATS Streaming maximum payload option to helm chart. The default is the same as the current one, 1Mb.

## Motivation and Context
It allows people who need to pass a variable to set the nats config max payload size. This allow to potentially do async request of more than 1Mb if configured, or to have a stricter limit.
Fixes https://github.com/openfaas/faas-netes/issues/734

## How Has This Been Tested?
Tested on a cluster setup with arkade and no costumisation, then updated with helm upgrade to test different payload size. Sent different payload to verify that the limit was correctly applied.
NB: you have to increase nginx max request size too if you dont want your request to be stopped before reaching nats.

## Types of changes

## Checklist:

### DIFF
--- a/chart/openfaas/README.md
+++ b/chart/openfaas/README.md
@@ -443,6 +443,7 @@ Additional OpenFaaS options in `values.yaml`.
 | `nats.external.enabled` | Whether to use an externally-managed NATS Streaming server | `false` |
 | `nats.external.host` | The host at which the externally-managed NATS Streaming server can be reached | `""` |
 | `nats.external.port` | The port at which the externally-managed NATS Streaming server can be reached | `""` |
+| `nats.maxPayload` | Setup the max payload size for NATS Streaming deployments managed by this chart | `1Mb` |
 | `nats.enableMonitoring` | Enable the NATS monitoring endpoints on port `8222` for NATS Streaming deployments managed by this chart | `false` |
 | `nats.metrics.enabled` | Export Prometheus metrics for NATS, no multi-arch support  | `false` |
 | `nats.metrics.image` | Container image used for the NATS Prometheus exporter, not multi-arch | `synadia/prometheus-nats-exporter:0.6.2` |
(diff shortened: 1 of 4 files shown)


CarbonC:
I didnt run make charts since it looks like to be likely to provoke conflicts. Tell me if anything else is needed!

alexellis:
My concern here is that you may be using NATS inappropriately where you should be using Minio or S3 to store your uploads before fetching them in a function.

Changing these kinds of default values can create unexpected side-effects and issues for us when we have to support this change for the community.

I would ask you to please elaborate on the problem you're trying to solve in the issue you raised.

Alex

CarbonC:
> My concern here is that you may be using NATS inappropriately where you should be using Minio or S3 to store your uploads before fetching them in a function.
> 
> Changing these kinds of default values can create unexpected side-effects and issues for us when we have to support this change for the community.
> 
> I would ask you to please elaborate on the problem you're trying to solve in the issue you raised.
> 
> Alex

Hi!
My problem here is that i have a lot of my requests, say 95%, that are under 1Mb, but a few are between 1 and 10Mb. Im not pushing really big things otherwise i would have used an external storage or split my payload. Not sure if a 2Mb request is really an issue...

CarbonC:
But more than my use case/the max_payload value, i feel like allowing to pass custom conf to parts of the stack and easily change them with configMaps is a good thing.

alexellis:
We will await your GitHub issue with the full template filled out.

albertkohl-monotek:
i second this! i'm not using helm, but i manually updated my .yml files and deployed it and it worked perfectly.

in my case, i'm using the async functions to pass a json string to send an email. that email can have attachments up to 20mb.

i know it's probably not best-use-case, but it works for me, and makes it so my site can offload the email to python scripts.

Issue #736, opened 2021-01-07
Only show openfaas Deployments in the API (regression)


pkoelbl:
## Expected Behaviour
The api call to system/functions should return only openfaas functions

## Current Behaviour
The api call to system/functions returns everything in namespaces set to be able to be used for faas functions.

## Steps to Reproduce (for bugs)
1. Set openfaas functions to be able to use a different namespace than openfaas-fn
2. deploy functions to the different namespace
3. deploy other services/deployments to the different namespace as well
4. make the api call to system/functions

## Context

All of our code is running in a particular namespace for organizational purposes. This includes faas functions that are deployed with faas-cli. Other services/deployments on that namespace include our website/api etc.
Using gateway 0.20.1 and faas-netes  0.12.8 the call to system/functions only returned the faas functions. We updated to 0.20.2 and 0.12.12 and now the call to system/functions includes everything.

This affected us because we made an azure queue listener that read from the annotations and the website/api doesn't have annotations so it caused an issue. We have worked around that to just ignore things with no annotations so it's not really a problem for us anymore but i figured i would report the change in behavior of the api.

## Your Environment

* FaaS-CLI version ( Full output from: `faas-cli version` ):
 commit:  c9d284d0c5bd90415baf9130fdce55af8e5a506b
 version: 0.12.19

* Docker version `docker version` (e.g. Docker 17.0.05 ):
20.10

* What version and distriubtion of Kubernetes are you using? `kubectl version`
Client Version: version.Info{Major:"1", Minor:"19", GitVersion:"v1.19.3", GitCommit:"1e11e4a2108024935ecfcb2912226cedeafd99df", GitTreeState:"clean", BuildDate:"2020-10-14T12:50:19Z", GoVersion:"go1.15.2", Compiler:"gc", Platform:"windows/amd64"}
Server Version: version.Info{Major:"1", Minor:"19", GitVersion:"v1.19.0", GitCommit:"e19964183377d0ec2052d1f1fa930c4d7575bd50", GitTreeState:"clean", BuildDate:"2020-08-26T18:49:11Z", GoVersion:"go1.15", Compiler:"gc", Platform:"linux/amd64"}

* Operating System and version (e.g. Linux, Windows, MacOS):

* Link to your project or a code example to reproduce issue:

* What network driver are you using and what CIDR? i.e. Weave net / Flannel

alexellis:
Hi @pkoelbl 

[...]

PR #738, opened 2021-01-08
Fix issue with filtering deployments

alexellis:
## Description

Fix issue with filtering deployments

## Motivation and Context

Fixes issue: #736 raised by pkoelbl where all deployments in
the openfaas-fn namespace were being returned via the REST
API when listing functions.

The selector's Add method was generating a new value to be
returned instead of updating itself. A negative test would have
shown this and may be worth adding to the codebase at a
later date, or being mocked/stubbed. A PR would be welcomed
for this.

## How Has This Been Tested?

I built a local image, downloaded `https://raw.githubusercontent.com/inlets/inlets-operator/master/contrib/nginx-sample-deployment.yaml`, updated its namespace and applied it into the functions namespace.

After deploying the new image, the nginx-1 deployment no-longer showed up on `faas-cli list`

Thanks also to @LucasRoesler for helping to investigate and confirm the fix.

## Types of changes

## Checklist:

If someone wanted to take on writing a test with a stub/fake that would be appreciated.


alexellis:
> If someone wanted to take on writing a test with a stub/fake that would be appreciated.

@LucasRoesler I would appreciate your help with this part

Issue #781, opened 2021-04-03
Unify helm chart values for multi-arch support


LucasRoesler:
Now that we publish all of our images as multi-arch to GHCR 

```sh
kubectl apply -f https://raw.githubusercontent.com/openfaas/faas-netes/master/namespaces.yml
helm upgrade openfaas --install openfaas/openfaas
```
should work on both arm and amd clusters (or even clusters with mixed node archs) without any special configuration. 

## Current Behaviour

We specify 

```yaml
nodeSelector:
  beta.kubernetes.io/arch: amd64
``` 

Which prevents the install command from working without special values, specifically we have to override the nodeSelector, for example

```yaml
nodeSelector:
  beta.kubernetes.io/arch: arm64
``` 

## Possible Solution

We should not specify the node arch via the `nodeSelector`, the same gateway image will work on all arm, amd64, and amd64, so we probably don't need to specify the selector at all.

Alternatively, if we want to really specify which nodes openfaas works with, we could use tolerations to create an explicit list of node types that we build multi-arch images for

```yaml
tolerations:
- key: beta.kubernetes.io/arch
  operator: Equal
  value: arm64
- key: beta.kubernetes.io/arch
  operator: Equal 
  value: amd64
- key: beta.kubernetes.io/arch
  operator: Equal
  value: arm
```

## Steps to Reproduce (for bugs)

1. create an ARM cluster
2. run the install 
```sh
kubectl apply -f https://raw.githubusercontent.com/openfaas/faas-netes/master/namespaces.yml
helm upgrade openfaas --install openfaas/openfaas
```
3. check that the pods are never actually scheduled
```sh
$ kubectl get po -n openfaas
NAME                                READY   STATUS    RESTARTS   AGE
alertmanager-7b864c9cb6-n9jnz       0/1     Pending   0          10s
prometheus-58c899f45d-xrjb7         0/1     Pending   0          10s
queue-worker-77f4446d48-wfbbl       0/1     Pending   0          10s
nats-7d86c64647-vzx5q               0/1     Pending   0          10s
basic-auth-plugin-bc899c574-r49ql   0/1     Pending   0          10s
gateway-7778bccd8-qx8zd             0/2     Pending   0          10s
```
Then describe one of the Pods and check the events
```
Events:
  Type     Reason            Age                From               Message
  ----     ------            ----               ----               -------
[...]

PR #782, opened 2021-04-04
Feat remove node selectors from helm chart

LucasRoesler:
## Description

- Remove the nodeSelector value from each of the values files in the
  helm chart. Now that we are using multi-arch builds for all of the
  services, the same configuration (specifically the same image name)
  is valid on any node type (amd64, arm64, and arm). As a result, we
  do not need to specify the node selector to ensure that the image is
  run on a compatible node, the node will automatically pull the correct
  variation of the image.
- This will simplify the integration of the microk8s addon for arm
  systems because the default installation will "just work"
- Reduce duplication of values in the values file for arm and armhf.
  This should make it easier to maintain because these settings are
  valid in any deployment. The values that remain in the arm and armhf
  files are the image name for the basic auth plugin, this one is not
  yet multi-arch, but once it is, we can remove this.
  The prometheus resources also request less memory so that the default
  fits better on a raspberry pi

## Motivation and Context

#781 

## How Has This Been Tested?

Use this command to render the chart on `master` and this branch, then used diff to detect any changes. For each values file, the only difference is the nodeSelector is now removed.  For example:

```sh
$ helm template ./chart/openfaas --values ./chart/openfaas/values.yaml --values ./chart/openfaas/values-arm64.yaml > arm64.yml
$ git checkout master
$ helm template ./chart/openfaas --values ./chart/openfaas/values.yaml --v
alues ./chart/openfaas/values-arm64.yaml > arm64-old.yml
$ diff arm64.yml arm64-old.yml
1378a1379,1380
>       nodeSelector:
>         beta.kubernetes.io/arch: arm64
1441a1444,1445
>       nodeSelector:
>         beta.kubernetes.io/arch: arm64
1585a1590,1591
>       nodeSelector:
>         beta.kubernetes.io/arch: arm64
1627a1634,1635
>       nodeSelector:
>         beta.kubernetes.io/arch: arm64
1699a1708,1709
>       nodeSelector:
>         beta.kubernetes.io/arch: arm64
1761a1772,1773
>       nodeSelector:
>         beta.kubernetes.io/arch: arm64
```

## Types of changes

## Checklist:

### DIFF
--- a/chart/openfaas/values.yaml
+++ b/chart/openfaas/values.yaml
@@ -54,7 +54,7 @@ oidcAuthPlugin:
 # Requires openfaasPRO=true
 # scale-to-zero feature
 faasIdler:
-  image: ghcr.io/openfaas/faas-idler-pro:0.4.2
+  image: ghcr.io/openfaas/faas-idler-pro:0.5.0
   replicas: 1
   create: true
   inactivityDuration: 30m               # If a function is inactive for 15 minutes, it may be scaled to zero
@@ -171,6 +171,7 @@ nats:
   image: nats-streaming:0.17.0
   enableMonitoring: false
   metrics:
+    # Should stay off by default because the exporter is not multi-arch (yet)
     enabled: false
     image: synadia/prometheus-nats-exporter:0.6.2
   resources:
--- a/chart/openfaas/values.yaml
+++ b/chart/openfaas/values.yaml
@@ -172,7 +172,7 @@ nats:
   enableMonitoring: false
   metrics:
     enabled: false
-    image: synadia/prometheus-nats-exporter:0.6.2 
+    image: synadia/prometheus-nats-exporter:0.6.2
   resources:
     requests:
       memory: "120Mi"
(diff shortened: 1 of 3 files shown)


chart/openfaas/values.yaml
```diff
@@ -171,8 +171,9 @@ nats:
   image: nats-streaming:0.17.0
   enableMonitoring: false
   metrics:
```
alexellis, line 173: Could you add this as explicitly off in the ARM values files?

Something like this, plus your comment should do the job?
```yaml
nats:
  enableMonitoring: false
```

alexellis (approved):
Approved

LucasRoesler:
@alexellis when will the next chart release be. If we do it soon, we can re-add the arm support notation to the microk8s addon before their next release, ie reverting this https://github.com/ubuntu/microk8s/pull/2142

LucasRoesler:
@alexellis RE tagging the release, the microk8s next release is v1.21 in a couple days (April 8) and then v1.22 is ~4 months after that. Per https://github.com/ubuntu/microk8s/pull/2142#issuecomment-814206175

So if we get a tag in today or tomorrow, then we might have some hope to get both ARM and AMD support on microk8s

PR #806, opened 2021-06-17
Add support for binary data in secrets

alexellis:

## Description

Add support for binary data in secrets

## Motivation and Context

Through the RawValue field in the latest version of the
faas-provider, users can now input raw binary data when
creating secrets.

Previously, we attempted to solve this problem by adding a 
`--trim=false` flag to faas-cli, but as @LucasRoesler pointed out, 
this isn't sufficient because of the use of StringData on the 
Kubernetes API.

This is not a breaking change in that:

* RawValue is a new and optional field
* StringData was only a helper method used by Kubernetes to populate its own Data field which is of type []byte

## How Has This Been Tested?

Unit tests updated.

```bash
arkade install openfaas --set faasnetes.image=ghcr.io/alexellis/faas-netes:binary-secrets-1 --set openfaasImagePullPolicy=Always
```

Followed by:

```bash
faas-cli secret create alex --from-literal is-here

kubectl get secret -n openfaas-fn alex -o jsonpath='{.data.alex}' | base64 --decode;echo

is-here
```

For binary data:

```bash
echo raw-data |base64
cmF3LWRhdGEK

curl -i -X POST http://admin:$PASSWORD@127.0.0.1:8080/system/secrets --data-binary '{"name": "binary1", "rawValue": "cmF3LWRhdGEK", "namespace": "openfaas-fn"}'

kubectl get secret -n openfaas-fn binary1 -o jsonpath='{.data.binary1}' | base64 --decode

raw-data
```

## Types of changes

## Checklist:

Once released, a change to the `faas-cli secret create` command will be needed to add a `--raw` flag.

### DIFF
--- a/pkg/k8s/secrets.go
+++ b/pkg/k8s/secrets.go
@@ -97,9 +97,16 @@ func (c secretClient) Create(secret types.Secret) error {
 				secretLabel: secretLabelValue,
 			},
 		},
-		StringData: map[string]string{
-			secret.Name: secret.Value,
-		},
+	}
+
+	if len(secret.RawValue) > 0 {
+		req.Data = map[string][]byte{
+			secret.Name: secret.RawValue,
+		}
+	} else {
+		req.Data = map[string][]byte{
+			secret.Name: []byte(secret.Value),
+		}
 	}
 
 	_, err = c.kube.Secrets(secret.Namespace).Create(context.TODO(), req, metav1.CreateOptions{})
--- a/Makefile
+++ b/Makefile
@@ -14,9 +14,10 @@ build-docker:
 
 .PHONY: build-buildx
 build-buildx:
-	@docker buildx create --use --name=multiarch --node=multiarch && \
+	@echo ghcr.io/$(OWNER)/faas-netes:$(TAG) && \
+	docker buildx create --use --name=multiarch --node=multiarch && \
 	docker buildx build \
-		--output "type=docker,push=false" \
+		--push \
 		--platform linux/amd64 \
 		--tag ghcr.io/$(OWNER)/faas-netes:$(TAG) \
 		.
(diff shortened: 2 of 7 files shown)


pkg/k8s/secrets.go
```diff
@@ -97,9 +97,16 @@ func (c secretClient) Create(secret types.Secret) error {
...
 			},
 		},
-		StringData: map[string]string{
-			secret.Name: secret.Value,
-		},
+	}
+
+	if len(secret.RawValue) > 0 {
+		req.Data = map[string][]byte{
+			secret.Name: secret.RawValue,
```
LucasRoesler, line 104: have you manually tested this, i know that when you get the value from the k8s API, these values are base64 encoded, so i don't know if we need to send them already encoded or if it accepts the non-encoded values during creation.

alexellis: I'll try that.

LucasRoesler: one other things, i think that we _need_ to require the `secret.RawValue` to be base64 encoded when it is submitted to faas-netes, this ensures that the payload is parsed correctly. Which means that either we need to decode it so that we can submit it raw  _or_ because it is already encoded, we wouldn't need to do anything and the code here would work as expected.

alexellis: Seems to work as expected for the existing path.

alexellis: I've done a couple of manual tests and updated the description. Let me know what you think.

alexellis:
I've resolved the review feedback and have also added comments to the existing Secret struct in faas-provider for the new and existing fields. 

Linked PR: https://github.com/openfaas/faas-cli/pull/895

Issue #841, opened 2021-10-06
OpenFaas Helm Chart not correctly detecting `PathType` needed


GerryWilko:
## Expected Behaviour

I am following the documentation for the [Kubernetes setup](https://docs.openfaas.com/reference/ssl/kubernetes-with-cert-manager/#add-tls-to-openfaas). I wish to deploy OpenFaaS gateway behind an `ingress-nginx` IngressController.

## Current Behaviour

At present I am unable to proceed past "Add TLS to OpenFaaS". It seems that the generated Kubernetes manifests are not adding `pathType` to the `host` on the Ingress YAML. Looking at the template it seems some work was done in July to detect the need for `pathType` https://github.com/openfaas/faas-netes/pull/817. If `pathType` is not supplied though there isn't a sensible default.

## List All Possible Solutions and Workarounds
I am raising a PR to correct the docs to specify `pathType`. Also perhaps in the `ingress.yaml` there should be a sensible default? `Prefix` is pretty typical and indeed before `pathType` existed it was the default.

## Which Solution Do You Recommend?

I would recommend a sensible default. `Prefix` should suffice for most use cases.

## Steps to Reproduce (for bugs)

1. Install `nginx-ingress`
2. Install OpenFaaS without a load balancer
3. Use the template `tls.yaml` values file from https://docs.openfaas.com/reference/ssl/kubernetes-with-cert-manager/#add-tls-to-openfaas
4. Observe unable to proceed on Kubernetes 1.19<

PR #842, opened 2021-10-06
Adding default pathType for Ingress #841

GerryWilko:
## Description

I have set a default `pathType` when `pathType` is necessary. I have set it to `Prefix`. Welcome feedback here but `Prefix` seems a sensible default unless otherwise specified with `.Values.ingress.pathType`.

## Motivation and Context

## How Has This Been Tested?

I have tested the chart now installs correctly when following the steps in #841 

## Types of changes

## Checklist:

- [ ] I have signed-off my commits with `git commit -s`


GerryWilko:
I've actually just noticed the comment in the `values.yaml` file that specify the need for `pathType`

> For k8s >= 1.18 you need to specify the pathType
> See https://kubernetes.io/blog/2020/04/02/improvements-to-the-ingress-api-in-kubernetes-1.18/#better-path-matching-with-path-types
>  pathType: ImplementationSpecific

Perhaps this should just be an update to the documentation? Though I don't see a problem with setting a sensible default. This shouldn't be a breaking change as previously the chart would error installing so users would have been unable to proceed. Let me know if instead you would just like the docs fixed and I'll happily change this PR to do that instead. 😄

LucasRoesler:
@GerryWilko can you share a copy of the error you were getting before? 

If we set a default, I think `ImplementationSpecific` would be a better default  given it is the specified default here https://kubernetes.io/blog/2020/04/02/improvements-to-the-ingress-api-in-kubernetes-1.18/#better-path-matching-with-path-types. 

Or alternatively,  using the empty value to simply omit it from the spec so that k8s can apply it's own default, but this is what it should already be doing

```yaml
            {{- if and $pathType $ingressSupportsPathType }}
            pathType: {{ $pathType }}
            {{- end }}
```

if pathType is empty, it omits the field in the rendered template and because k8s already has a default, it shouldn't cause an error

alexellis:
@GerryWilko in the meantime, have you tried arkade yet? It is Kubernetes version aware:

```bash
arkade install openfaas-ingress --help
```

GerryWilko:
@alexellis I'm thinking of using it on two projects. One for syncing and processing of accelerometer data. Another for an open source configuration management tool.

I really hate the idea of doing things on cloud provider specific services so OpenFaaS is the answer I hope!

PR #845, opened 2021-10-11
Migrate OIDC 0.5.0

alexellis:
## Description

Migrate to OIDC plugin 0.5.0

## Motivation and Context

    * Renames the deployment for the plugin to match its name
    better (oidc plugin, not oauth2-plugin)
    * Removes the need for separate token/autorization URLs
    by adding an openid discovery URL
    * Updates to version 0.5.0 of the component
    * Reads license from a secret in the cluster instead of a literal
       value. Thanks @dirien for the suggestion
   
## How Has This Been Tested?

    Tested e2e with the code grant flow and Auth0/Google as the
    sign-on provider.

## Types of changes

Existing users will need to redeploy the helm chart if they are using the SSO capability.

### DIFF
--- a/chart/kafka-connector/README.md
+++ b/chart/kafka-connector/README.md
@@ -36,7 +36,7 @@ The [Kafka connector](https://github.com/openfaas-incubator/kafka-connector) bri
 $ kubectl create secret generic \
     -n openfaas \
     openfaas-license \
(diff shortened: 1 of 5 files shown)


alexellis:
The accompanying docs change: https://github.com/openfaas/docs/commit/4f9a7b068c7624195f7f6fbd988448fd7ad2bda0

Issue #868, opened 2021-10-30
Gateway Pod crashes when Profiles CRD is deleted


kevin-lindsay-1:
## My actions before raising this issue
- [ ] Followed the [troubleshooting guide](https://docs.openfaas.com/deployment/troubleshooting/)
- [ ] Read/searched [the docs](https://docs.openfaas.com/)
- [ ] Searched [past issues](/issues)

@alexellis Posting here and plan to edit and follow-up, so that this issue doesn't get lost.

## Expected Behaviour

I don't think that this should crash the gateway, but I do think an error should be logged.

## Current Behaviour

Gateway crashes if you try to create a function with a profile; this seems initially related to using the operator and CRD flags set to `true` on the `faas-netes/openfaas` helm chart.

## Are you a GitHub Sponsor (Yes/No?)

Check at: https://github.com/sponsors/openfaas
- [ ] Yes
- [ ] No
- [x] No, but I sponsor Alex

## Context

Crashing our gateway instances rather than simply not creating functions.

* FaaS-CLI version ( Full output from: `faas-cli version` ):
0.13.13

* Docker version `docker version` (e.g. Docker 17.0.05 ):
20.10.8

* Which deployment method do you use?:
- [x] OpenFaaS on Kubernetes
- [ ] [faasd](https://github.com/openfaas/faasd)

server v1.21.3
client v1.22.2

* Operating System and version (e.g. Linux, Windows, MacOS):
MacOS

* Code example or link to GitHub repo or gist to reproduce problem:

* Other diagnostic information / logs from [troubleshooting guide](https://docs.openfaas.com/deployment/troubleshooting)

## Next steps

You may [join Slack](https://docs.openfaas.com/community) for community support.

alexellis:
The gateway itself won't crash due to profiles since it doesn't have anything to do with K8s, this is likely the operator container in the same pod. I'm moving this over to the faas-netes repo so someone can look into it.

alexellis:
@LucasRoesler please could you take a look into this?

alexellis:
If anyone else should get here first, the chart has to be installed with operator.create to match Kevin's setup. Here's how to define a profile -> https://docs.openfaas.com/reference/profiles/

The K8s version Kevin's using is 1.21, but this may also be an issue with changes in 1.22?

Install with helm or arkade.

LucasRoesler:
@alexellis taking a look now

cpanato:
[...]

PR #872, opened 2021-11-06
fix: improve error messages when teh profile crd is missing

LucasRoesler:
Add explicit errors during startup and during deploy when the Profiles
CRD is missing. This will not prevent deployment and invocation of functions that don't require Profiles but it will block functions that use the feature.

This improves the error behavior and makes the errors more explicit so that it is easier for the cluster operator to fix it.

## Motivation and Context

Relates to #868

## How Has This Been Tested?

WIP

## Types of changes

## Checklist:

### DIFF
--- a/main.go
+++ b/main.go
@@ -90,6 +90,14 @@ func main() {
 		log.Fatalf("Error reading config: %s", err.Error())
 	}
 
+	hasProfiles, err := k8s.ProfilesEnabled(kubeClient)
+	if err != nil {
+		log.Fatalf("Error getting preferred APIs: %s", err.Error())
+	}
+	if !hasProfiles {
+		log.Fatalf("Profiles CRD not found, disabling the Profiles feature. Warning, functions that request a Profile will be unschedulable.")
+	}
+
 	config.Fprint(verbose)
 
 	deployConfig := k8s.DeploymentConfig{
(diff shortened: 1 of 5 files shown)


pkg/handlers/deploy.go
```diff
@@ -42,10 +42,22 @@ func MakeDeployHandler(functionNamespace string, factory k8s.FunctionFactory) ht
 			defer r.Body.Close()
 		}
 
+		hasProfiles, err := k8s.ProfilesEnabled(factory.Client)
```
alexellis, line 45: Previously, would this handler work if the CRD wasn't present and we had no annotation for profiles in the request?

pkg/controller/controller.go
```diff
@@ -301,6 +311,15 @@ func (c *Controller) syncHandler(key string) error {
 			return err
 		}
 
+		hasProfiles, err := k8s.ProfilesEnabled(c.kubeclientset)
+		if err != nil {
+			return err
+		}
+
+		if !hasProfiles {
```
alexellis, line 319: If this crashes when the CRD isn't present, how does the code get to the handlers for update/deploy? Would this be helpful enough on its own?

main.go
```diff
@@ -90,6 +90,14 @@ func main() {
 		log.Fatalf("Error reading config: %s", err.Error())
 	}
 
+	hasProfiles, err := k8s.ProfilesEnabled(kubeClient)
+	if err != nil {
+		log.Fatalf("Error getting preferred APIs: %s", err.Error())
+	}
+	if !hasProfiles {
```
alexellis, line 97: If this crashes when the CRD isn't present, how does the code get to the handlers for update/deploy? Would this be helpful enough on its own?

pkg/handlers/deploy.go
LucasRoesler: yes, if there are no profile annotations present, then `profileList, err = factory.GetProfiles` on L90 should return an empty list as a noop, there are various spots in the profiles code that will just be empty loops.

main.go
LucasRoesler: the profiler CRD can be deleted _before_ startup, which this code catches, or _after_ startup, which the code in the handlers then catches.

PR #978, opened 2022-07-01
Instrument the HTTP endpoints of the provider

alexellis:

## Description

Instrument the HTTP endpoints of the provider

## Motivation and Context

After adding Prometheus metrics to faas-provider, users will
be able to collect metrics from faas-netes.

At the same time, the project is migrated to Go 1.18

## How Has This Been Tested?

Tested by port-forwarding the gateway pod and accessing the
new /metrics path on the preexisting HTTP port of the
container.

The helm chart and Prometheus config changes are reflected as follows:

Screenshot from 2022-07-01 16-14-33

## Types of changes

## Checklist:

This new data can be added to the metrics page in the openfaas docs.

### DIFF
--- a/chart/openfaas/templates/prometheus-cfg.yaml
+++ b/chart/openfaas/templates/prometheus-cfg.yaml
@@ -39,38 +39,37 @@ data:
         static_configs:
           - targets: ['localhost:9090']
 
-      - job_name: "openfaas-services"
-        scrape_interval: 5s
-        honor_labels: false
+      # Capture endpoints in the openfaas namespace with a scrape annotation
+      # such as the gateway-provider service.
+      - job_name: 'openfaas-endpoints'
         kubernetes_sd_configs:
-          - role: service
-            namespaces:
-              names:
-                - {{ .Release.Namespace }}
+        - role: endpoints
+          namespaces:
+            names:
+              - {{ .Release.Namespace }}
         relabel_configs:
         - action: labelmap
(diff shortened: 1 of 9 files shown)


chart/openfaas/templates/prometheus-cfg.yaml
```diff
@@ -39,6 +39,39 @@ data:
         static_configs:
           - targets: ['localhost:9090']
 
+      - job_name: "openfaas-services"
```
alexellis, line 42: This isn't meant to be here.

chart/openfaas/templates/prometheus-cfg.yaml
```diff
@@ -39,6 +39,38 @@ data:
         static_configs:
           - targets: ['localhost:9090']
 
+      # Capture endpoints in the openfaas namespace with a scrape annotation
+      # such as the gateway-provider service.
+      - job_name: 'openfaas-endpoints'
```
alexellis, line 44: @LucasRoesler to make this change I went for a combination of a couple of your suggestions.

1) New service for the provider, limited in scope to only a ClusterIP
2) Endpoint rather than Service SD for K8s

Then I restricted this further by only matching where the scrape and port annotations were present.

LucasRoesler: Looks good, I'm glad the research made sense and was helpful

LucasRoesler (dismissed):

PR #1013, opened 2022-09-02
expose gateway probes

kevin-lindsay-1:
Exposes the readiness and liveness probe timing & success/failure settings to be able to be set by the user.

## Description

Pod use readiness for load balancing and preventing pod overloading. Liveness determines if a pod has gotten "stuck", and restarts it if it stops responding.

## Motivation and Context

## How Has This Been Tested?

Local deployment of chart.

## Types of changes

- [X] New feature (non-breaking change which adds functionality)

## Checklist:

### DIFF
--- a/chart/openfaas/templates/gateway-dep.yaml
+++ b/chart/openfaas/templates/gateway-dep.yaml
@@ -64,6 +64,13 @@ spec:
           runAsUser: 10001
         {{- end }}
         livenessProbe:
+          {{- if $.Values.gateway.livenessProbe }}
+          initialDelaySeconds: {{ $.Values.gateway.livenessProbe.initialDelaySeconds | default 0 }}
+          periodSeconds: {{ $.Values.gateway.livenessProbe.periodSeconds | default 10 }}
+          timeoutSeconds: {{ $.Values.gateway.livenessProbe.timeoutSeconds | default 5 }}
+          failureThreshold: {{ $.Values.gateway.livenessProbe.failureThreshold | default 3 }}
+          successThreshold: {{ $.Values.gateway.livenessProbe.successThreshold | default 1 }}
+          {{- end }}
           {{- if .Values.httpProbe }}
           httpGet:
             path: /healthz
@@ -78,8 +85,14 @@ spec:
             - --spider
             - http://localhost:8080/healthz
           {{- end }}
-          timeoutSeconds: 5
         readinessProbe:
+          {{- if $.Values.gateway.readinessProbe }}
+          initialDelaySeconds: {{ $.Values.gateway.readinessProbe.initialDelaySeconds | default 0 }}
+          periodSeconds: {{ $.Values.gateway.readinessProbe.periodSeconds | default 10 }}
+          timeoutSeconds: {{ $.Values.gateway.readinessProbe.timeoutSeconds | default 5 }}
+          failureThreshold: {{ $.Values.gateway.readinessProbe.failureThreshold | default 3 }}
+          successThreshold: {{ $.Values.gateway.readinessProbe.successThreshold | default 1 }}
+          {{- end }}
           {{- if .Values.httpProbe }}
           httpGet:
             path: /healthz
@@ -94,7 +107,6 @@ spec:
             - --spider
             - http://localhost:8080/healthz
           {{- end }}
-          timeoutSeconds: 5
         env:
         {{- if .Values.gateway.logsProviderURL }}
         - name: logs_provider_url
--- a/chart/openfaas/values.yaml
+++ b/chart/openfaas/values.yaml
@@ -132,6 +132,20 @@ gateway:
       memory: "120Mi"
       cpu: "50m"
 
+  readinessProbe:
+    initialDelaySeconds: 0
+    periodSeconds: 10
+    timeoutSeconds: 5
+    failureThreshold: 3
+    successThreshold: 1
+
+  livenessProbe:
+    initialDelaySeconds: 0
+    periodSeconds: 10
+    timeoutSeconds: 5
+    failureThreshold: 3
+    successThreshold: 1
+
 basicAuthPlugin:
   image: ghcr.io/openfaas/basic-auth:0.23.2
   replicas: 1


chart/openfaas/templates/gateway-dep.yaml
```diff
@@ -78,8 +83,12 @@ spec:
             - --spider
             - http://localhost:8080/healthz
           {{- end }}
-          timeoutSeconds: 5
         readinessProbe:
```
alexellis, line 86: What would happen for people who have a values.yaml file that they maintain separately from the main chart?

Do we need some kind of defaulting or fallback? cc @welteki

kevin-lindsay-1: What do you mean? If someone provides a value at `gateway.livenessProbe.X`, it shouldn't do anything in the current chart, so I have no idea why they'd put a value there.

If they're altering the templates or something provided by this chart, that's "here be dragons" territory, where support wouldn't really exist.

alexellis: What do I mean?

What would happen here?

```
       livenessProbe:
           initialDelaySeconds: {{ .Values.gateway.livenessProbe.initialDelaySeconds }}
```

If there was someone who had their own values file and didn't have this new section you've added, called `gateway.livenessProbe`? I suspect "0" or "" would be inserted, or maybe it would error.

Whenever we add fields, we try to make sure it won't break for any users.

So I'd suggest to get this merged, we need some form of or and default value that is the same as what you get prior to this PR.

alexellis (changes requested):
This needs some changes to support existing users.

alexellis:
@welteki please can you make a suggestion for @kevin-lindsay-1 or point him to one of your prior examples with defaulting / if / or statements?

welteki:
Hi Kevin, like Alex mentioned you can use `if` statements, you can also use the default function like we do in several other places in the chart.

https://github.com/openfaas/faas-netes/blob/85d425bfae93d00a4a053d874418b8ff32aeaf80/chart/openfaas/templates/ingress.yaml#L3
https://github.com/openfaas/faas-netes/blob/85d425bfae93d00a4a053d874418b8ff32aeaf80/chart/openfaas/templates/gateway-dep.yaml#L2

https://helm.sh/docs/chart_template_guide/functions_and_pipelines/#using-the-default-function

chart/openfaas/templates/gateway-dep.yaml
kevin-lindsay-1: Oh, I see what you mean. If someone updates the chart but doesn't update their values file, too. Makes sense.

kevin-lindsay-1:
That linting error I see seems like the linter is supplying a type that I wouldn't expect to see in a real chart?

chart/openfaas/templates/gateway-dep.yaml
kevin-lindsay-1: I went with the `dig | default` pattern that I use in many charts and templates. Clean and easy way to template values, using what is effectively optional chaining. It seems that the linter doesn't like it, though?

kevin-lindsay-1: If `.Values.favorites` doesn't exist, the template will get a nil pointer exception when trying to access `.drink`

kevin-lindsay-1: Here's an example: https://go.dev/play/p/tSsm8RoFqIN

alexellis:
```
Error:  templates/: template: openfaas/templates/gateway-dep.yaml:67:45: executing "openfaas/templates/gateway-dep.yaml" at <dig "gateway" "livenessProbe" "initialDelaySeconds" nil>: error calling dig: interface conversion: interface {} is chartutil.Values, not map[string]interface {}
```

Perhaps we need a solution that avoids dig?

If gateway.livenessProbe => then try to read the whole path, otherwise set the default.

chart/openfaas/templates/gateway-dep.yaml
kevin-lindsay-1: Helm uses sprig v3, which is stated in their docs.

https://github.com/helm/helm/blob/main/pkg/engine/funcs.go

They seem to override a function or two from sprig with a helm variant, but I think the contract of those functions is effectively the same.

kevin-lindsay-1:
Updated to not dig `$.Values` directly, since its type is not a `map[string]interface{}`.

alexellis:
I'm seeing:

> Error:  templates/: parse error at (openfaas/templates/gateway-dep.yaml:67): function "dig" not defined

kevin-lindsay-1:
That appears to be due to sprig functions not being in scope in whatever linter is being used, which effectively means that it isn't linting according to helm's contract. Helm has dig.

alexellis:
Values.gateway has always been there, so we only need to check to Values.gateway.livenessProbe IMHO, that should be possible without the dig macro? If Values.gateway.livenessProbe is defined and empty, then that's a user error, not an upgrade issue.

kevin-lindsay-1:
So if a user replaces `livenessProbe` with `{}`, you want to not include them, or use your preferred default values?

kevin-lindsay-1:
If you don't want to use dig, ok. It looks like the lint issue is an issue with your builder, but ok. I'll do whatever, because this isn't my repo. If you want me to write an if statement and have a separate livenessProbe with defaults, I'll do it. I don't think it's clean, but I'm not going to die on a hill for this PR, which I was doing as a favor.

alexellis:
> So if a user replaces livenessProbe with {}, do you want to not include them, or do you want it to use your preferred default values?

We'll leave this as an unsupported use-case.

If a user wants K8s defaults, they can copy them into livenessProbe

chart/openfaas/templates/gateway-dep.yaml
```diff
@@ -78,8 +91,20 @@ spec:
...
+          {{- if $.Values.gateway.readinessProbe }}
+          initialDelaySeconds: {{ $.Values.gateway.readinessProbe.initialDelaySeconds | default 0 }}
+          periodSeconds: {{ $.Values.gateway.readinessProbe.periodSeconds | default 10 }}
+          timeoutSeconds: {{ $.Values.gateway.readinessProbe.timeoutSeconds | default 5 }}
+          failureThreshold: {{ $.Values.gateway.readinessProbe.failureThreshold | default 3 }}
+          successThreshold: {{ $.Values.gateway.readinessProbe.successThreshold | default 1 }}
+          {{- else }}
```
alexellis, line 101: Kevin mentioned: else could be omitted.

kevin-lindsay-1: done

alexellis (approved):
LGTM

alexellis:
@kevin-lindsay-1 the chart has been republished with this change. Thanks for working on it.

PR #1031, opened 2022-10-04
Feat add startup error when running in kube system

LucasRoesler:
## Description

To avoid any accedential security issues, we block running anything in
the kube-system namespace. We already have this explicitly blocked in
the rest of the code that deals with namespaces and it causes hard to
debug errors for users that try to deploy to the kube-system namespace.

This adds an explicit check so that this misconfiguration is easier to
detect and debug for end users.
## Motivation and Context

Part of a comment on #1022

## Types of changes

## Checklist:


main.go
```diff
@@ -98,6 +98,12 @@ func main() {
 
 	config.Fprint(verbose)
 
+	// use kubeclient to check the current namespace
+	namespace, _ := k8s.CurrentNamespace()
+	if namespace == "kube-system" {
```
alexellis, line 103: This isn't quite the error we were running into.

You probably can run the openfaas core components in the kube-system namespace.

It was that you can't deploy functions there.

So we already have validation in the deployment handlers, but we could do better by checking the functionNamespace variable (if set), and exiting upon startup

pkg/k8s/namespaces.go
```diff
@@ -0,0 +1,27 @@
...
+)
+
+// CurrentNamespace attempts to return the current namespace from the environment
+// or from the service account file. If it cannot find the namespace, it returns
+// an empty string. This will be empty when the not running in-cluster.
+//
+// This implementation is based on the clientcmd.inClusterClientConfig.Namespace method.
+// This is not exported and not accessible via other methods, so we have to copy it.
+func CurrentNamespace() (namespace string, found bool) {
+	if ns := os.Getenv("POD_NAMESPACE"); ns != "" {
```
alexellis, line 16: That is a pretty nice trick. TIL

Although please see my other command

main.go
LucasRoesler: i thought this was part of the suggestion https://github.com/openfaas/faas-netes/issues/1022#issuecomment-1253906489

but i see that what we also need to do is provide a meaningful api error as well

PR #1111, opened 2023-06-06
Remove check for openfaasPro for operator in Helm template

alexellis:
## Description

Remove check for openfaasPro for operator

## Why is this needed?

There is no need to check for openfaasPro when operator.create is set to true, because it's only available for openfaasPro customers.

This is some additional optimisation from previous commits.

## How Has This Been Tested?

The following gave no errors:

```
helm template chart/openfaas --set operator.create=true
helm template chart/openfaas --set operator.create=true --set openfaasPro=true
```

## Types of changes

Refactoring

### DIFF
--- a/chart/openfaas/Chart.yaml
+++ b/chart/openfaas/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 type: application
 description: OpenFaaS - Serverless Functions Made Simple
 name: openfaas
-version: 13.0.0
+version: 13.0.1
 sources:
 - https://github.com/openfaas/faas
 - https://github.com/openfaas/faas-netes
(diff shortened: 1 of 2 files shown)


alexellis:
That's the intention and is already the case prior to this PR. 

I'm not sure what else you'd suggest?

welteki:
> That's the intention and is already the case prior to this PR.

Previously the helm deploy would succeed but the `operator` pod would not start. Checking the logs would tell users a license is required.

With this change the helm deploy fails with the above mentioned error which might not be clear for users.

> I'm not sure what else you'd suggest?

We could make it fail with a custom message and make the cause of the failure clear to users.

```
{{- if and .Values.operator.create (not .Values.openfaasPro) }}
  {{- fail "enabling 'operator.create' is only supported for OpenFaaS Pro" }}
{{- end }}
```

alexellis:
I'll give that a try - thanks. TIL "fail" is a thing for Helm

alexellis:
@welteki do you want to send your own PR that applies the spirit of my change, but with the "fail" approach?

PR #1144, opened 2023-08-31
Support setting the backoff algorithm for retries

welteki:
## Description

Support selecting different modes for the exponential backoff algorithm. The value must be one of `exponential`, `full` of `equal`.

```yaml
queueWorkerPro:
  backoff: "full"
```

## Why is this needed?

An option was added to the queue-worker to select different jitter modes for the exponential backoff algorithm. This change allows configuring these modes through the helm chart.

## How Has This Been Tested?

This version of the chart was used to deploy the queue-worker while testing the backoff mode feature.

## Types of changes

## Checklist:

### DIFF
--- a/chart/openfaas/templates/jetstream-queueworker-dep.yaml
+++ b/chart/openfaas/templates/jetstream-queueworker-dep.yaml
@@ -79,6 +79,8 @@ spec:
           value: "{{ .Values.jetstreamQueueWorker.logs.format }}"
         - name: upstream_timeout
           value: "{{ .Values.gateway.upstreamTimeout }}"
+        - name: backoff
+          value: "{{ .Values.queueWorkerPro.backoff }}"
 
         # OpenFaaS PRO license required
         - name: "max_retry_attempts"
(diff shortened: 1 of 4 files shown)


chart/openfaas/templates/jetstream-queueworker-dep.yaml
```diff
@@ -79,6 +79,8 @@ spec:
           value: "{{ .Values.jetstreamQueueWorker.logs.format }}"
         - name: upstream_timeout
           value: "{{ .Values.gateway.upstreamTimeout }}"
+        - name: backoff
```
alexellis, line 82: What happens if empty string is inserted instead, i.e. just whatever we see in `.Values.queueWorkerPro.backoff`?

Do we not need a change to the dedicated queue-worker chart too?

welteki: The queue-worker will use a default of `exponential` when the env variable is empty. I guess we could leave out the default.

welteki: This has now been updated and the backoff value has also been added to the dedicated queue-worker chart.

alexellis (approved):
Approved

PR #1253, opened 2026-06-23
Add allowedCallbackURLs parameter for queue-worker

welteki:
## Description

Allow users to restrict which hosts receive async function callbacks via a list of glob patterns. The `allowedCallbackURLs` parameter is added to both the main openfaas chart and the standalone queue-worker chart.

- Defaults to `["*"]` for backwards compatibility (allow all)
- Set to empty list `[]` to deny all callbacks
- Accepts a list of URLs like `*.example.com` or `api.internal`

## Why is this needed?

Prevents SSRF-style attacks where malicious callback URLs could be used to reach internal services. Users can now restrict callbacks to trusted domains only.

- [x] I have raised an issue to propose this change

## How Has This Been Tested?

This chart was used and validated during E2E testing.

## Types of changes

## Checklist:


welteki:
> Finding: Helm chart changes are not versioned for release

This is intentional. We bump the version before release as multiple PRs with changes may be bundeled in a single release.

